PR #23, opened 2021-08-13
Add missing function 20 execution

santoshpuranik:
Execution call for function 20 is missing after integrating functions 14
through 19. Add it back.

Change-Id: I2354fd6f21dafdeb68c33f83ff0bbbb748563213
Signed-off-by: Santosh Puranik <santosh.puranik@in.ibm.com>


geissonator:
add to approvelist

PR #48, opened 2021-10-09
Bug fix for TEST cases

GiridhariKrishnan:
Changes with bug fixes and for Gtest compatability
with openBMC CI for State Manager TEST cases.


geissonator:
In order to get CI, need to do PR's against the 1020 branch (not main).

include/panel_state_manager.hpp
```diff
@@ -141,6 +140,13 @@ class PanelStateManager
     /** @brief Api to display function 2 on panel */
     void displayFunc02() const;
 
+    /**
```
jinuthomas, line 143: why this addition ?

include/panel_state_manager.hpp
```diff
@@ -84,10 +80,13 @@ class PanelStateManager
 
     /**
      * @brief Constructor.
-     * @param[in] transport - transport object to call transport functions.
+     * @param[in] transport - transport object to call transport functions
+     * @param[in] execute - pointer to executor class.
      */
-    PanelStateManager(std::shared_ptr<Transport> transport) :
-        transport(transport)
+    PanelStateManager(std::shared_ptr<Transport> transport,
```
jinuthomas, line 86: why this change is not in your Gerrit ?

GiridhariKrishnan: I think the changes are merged in Gerrit but not here, so the difference. I'll try to change that.

include/panel_state_manager.hpp
```diff
@@ -191,6 +197,9 @@ class PanelStateManager
 
     /*shared pointer to executor object*/
     std::shared_ptr<Executor> funcExecutor;
+
```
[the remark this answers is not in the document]
GiridhariKrishnan: These are merged in Gerrit but not in this. I'll check over which patchiest this to be based to.

jinuthomas (changes requested):
why are there two commits , please create a fresh PR based on master , why would Gerri and this also differ, it should not, I always integrate here first then Gerri. please reconfirm.

PR #55, opened 2021-11-02
Panel function state toggle.

SunnySrivastava1984:
This commit implement changes required to set state of panel functions depending upon various system parameters.
It also provides Dbus api toggle panel function state.


jinuthomas (approved):
Looks OK to me

santoshpuranik (approved):

PR #75, opened 2021-12-07
Panel function 74 implementation

SunnySrivastava1984:
The function on execution triggers a timer for a given
duration and sets Dbus property to mark window to upload
ACF certificate as active.
This property should be looked upon by Bmcweb to detect
if the window has been enabled.

Upon timer expiration the window is set as disabled.

Change-Id: I51a890b48c38fc52d94ff336a469f89c95f26160


joseph-reynolds (approved):
I am not familiar with ibm-panel code, boost::asio timers, or sdbusplus coding, but this seems okay to me.
As far as I can tell, when the BMC reboots or the application restarts: ACFWindowActive=false.  That behavior seems correct.
When function74 is invoked or re-invoked: it sets or resets the 30 minute window with ACFWindowActive=true, and then after the timeout it goers back to ACFWindowActive=false.  That behavior seems correct.

joseph-reynolds:
How can I activate this function for testing?   What is the D-bus API to invoke OpPanel function74?  Or can I use a D-Bus API to directly set ACFWindowActive=true or false?

SunnySrivastava1984:
> 



> I am not familiar with ibm-panel code, boost::asio timers, or sdbusplus coding, but this seems okay to me. As far as I can tell, when the BMC reboots or the application restarts: ACFWindowActive=false. That behavior seems correct. When function74 is invoked or re-invoked: it sets or resets the 30 minute window with ACFWindowActive=true, and then after the timeout it goers back to ACFWindowActive=false. That behavior seems correct.

Sounds good.

SunnySrivastava1984:
> How can I activate this function for testing? What is the D-bus API to invoke OpPanel function74? Or can I use a D-Bus API to directly set ACFWindowActive=true or false?

You need to have panel hardware, using which you need to traverse to function 74 and press an execute on that.

joseph-reynolds:
> > How can I activate this function for testing? What is the D-bus API to invoke OpPanel function74? Or can I use a D-Bus API to directly set ACFWindowActive=true or false?
> 
> You need to have panel hardware, using which you need to traverse to function 74 and press an execute on that.

Sorry I phrased my question wrong.  I am familiar with the operation of the physical panel, but that was not my question.

My question is: For unit testing, I will not have access to the OpPanel; I want a software solution to simulate pressing function 74 and I though I heard there was a way to do that (given BMC shell access as root).  How can I simulate pressing function 74?

jinuthomas (approved):

PR #101, opened 2022-02-16
Revert "Get OS IPL mode state from PHYP"

PriyangaRamasamy:
This reverts commit 90f621ea717d5f35e8878a4d2cd0078a408d81f5.

Reason for revert: Panel app exits due to
sensor PDR fetch failure. This has happened
before the PLDM service starts. Need to handle
this case. So reverting the commit back.

Change-Id: I3423017152d54ded00b354c4794e4203f4847ae8


santoshpuranik:
Not needed anymore

PR #130, opened 2022-04-12
Fix:SW547181 Display static/DHCP IP if present

PriyangaRamasamy:
This commit fixes function 30 defect in ibm-panel that
was overriding static/dhcp ip with link local when the
mac address was not found at that point of time.

This is fixed in such a way that once the static/dhcp ip is
found, then function 30 won't parse any other network
objects.

Test:
Tested on rainier.
Recreated the same issue and tested for eth1 where it has both
static and linklocal and with this static ip was displayed on panel
as expected.
output:
L1 : SP: ETH1:      T1
L2 : 9.3.29.29

Change-Id: I54a23c61f0d44c2811bbe4ebef7ea7f1b30db31d

### DIFF
--- a/src/executor.cpp
+++ b/src/executor.cpp
@@ -368,11 +368,9 @@ void Executor::execute30(const types::FunctionalityList& subFuncNumber)
     std::string ethObjPath = constants::networkManagerObj;
     ethObjPath += "/";
     ethObjPath += ethPort;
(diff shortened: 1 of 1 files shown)


jinuthomas:
jenkins run tests please

santoshpuranik:
jenkins run tests please

PR #148, opened 2022-07-13
transport: Recover From Bootloader Hang

santoshpuranik:
This commit implements a workaround for an issue seen with certain
versions of the panel microcode's bootloader. Due to a bug, the
bootloader can sometimes get stuck and not handover to the main panel
application.

This renders the panel unusable (applies to both LCD and bootfail
microcode). I2C to the PIC still works and the bootloader has the
capability to jump to the main program if sent the right I2C command:
0xFF25...

This commit checks if the PIC is stuck in the bootloader whenever we
enable the transport class for the respective PIC. It reads the version
off of the PIC. The bootloader returns a unique version string: "BL".

In case we find the PIC is stuck, we send it the I2C command to jump
over to the main program.

Tested: Tested on a Rainier system by forcing the PIC into the
bootloader and then running the panel app. Verified that the recovery
succeeds and we are able to get the PIC to jump to the main program.

Signed-off-by: Santosh Puranik <santosh.puranik@in.ibm.com>

### DIFF
--- a/include/i2c_message_encoder.hpp
+++ b/include/i2c_message_encoder.hpp
@@ -105,11 +105,18 @@ class MessageEncoder
     Binary softReset();
 
     /** @brief Flash Update command encode api
-     * The encoded data packet contains the command code of flash update(0xFF3)
+     * The encoded data packet contains the command code of flash update(0xFF30)
      * and the calculated checksum.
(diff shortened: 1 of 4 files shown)


jinuthomas (approved):

geissonator:
add to approvelist

PR #155, opened 2022-08-04
Display default function after panel hotplug

PriyangaRamasamy:
After a panel hotplug, lcd panel now displays default
function 01. This gives a visibility to the user that
the panel is up again after a hotplug.

Test:

Tested for both the cases where BMC is NotReady & Ready.

=> When BMC is set to NotReady, and if the lcd panel is present,
we don't get the function 01 display.
Then when BMC is set to Ready, (lcd panel is still present),
function 01 execution takes place (from the BMCState Properties
changed callback).

=> When BMC is Ready and if Panel's present property is changed
from false to true, function 01 is executed.

=======xxx========

busctl set-property xyz.openbmc_project.Inventory.Manager /xyz/openbmc_project/inventory/system/chassis/motherboard/lcd_op_panel_hill xyz.openbmc_project.Inventory.Item  Present b false
busctl set-property xyz.openbmc_project.Inventory.Manager /xyz/openbmc_project/inventory/system/chassis/motherboard/lcd_op_panel_hill xyz.openbmc_project.Inventory.Item  Present b true

./ibm-panel
Transport key is set to 0 for the panel at /dev/i2c-7, Z

Failed to determine OR fix bootloader bug ...

 Panel:Soft reset done.

 Button configuration done.

Transport key is set to 1 for the panel at /dev/i2c-7, Z

L1 : 01     N    PVM
L2 :             P

Change-Id: I159316f911dfaba3174026fa4689b48111f8c1fa

### DIFF
--- a/include/bus_monitor.hpp
+++ b/include/bus_monitor.hpp
@@ -31,9 +31,10 @@ class PanelPresence
      */
     PanelPresence(std::string& objPath,
                   std::shared_ptr<sdbusplus::asio::connection> conn,
-                  std::shared_ptr<Transport> transport) :
+                  std::shared_ptr<Transport> transport,
+                  std::shared_ptr<state::manager::PanelStateManager> state) :
         objectPath(objPath),
-        conn(conn), transport(transport)
+        conn(conn), transport(transport), stateManager(state)
     {
     }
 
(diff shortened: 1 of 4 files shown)


jinuthomas (changes requested):
Is this needed for 1020.10 ?

jinuthomas:
Not required for 1020.10

PR #156, opened 2022-08-04
Panel CM in Everest

SunnySrivastava1984:
The commit make changes related to panel CM in case of
Everest system.
As a process of which in case base panel is detached and
re-attached state of all the LEDs on base panel needs to
be restored.


src/bus_monitor.cpp
```diff
@@ -10,6 +10,77 @@
...
+        auto res = utils::readBusProperty<std::variant<bool>>(
+            "xyz.openbmc_project.LED.GroupManager", led,
+            "xyz.openbmc_project.Led.Group", "Asserted");
+
+        if (auto asserted = std::get_if<bool>(&res))
+        {
+            if (*asserted)
+            {
+                std::cout << "Led asserted = " << led << std::endl;
+                utils::writeBusProperty<bool>(
```
santoshpuranik, line 35: These can throw - and I don't think we should let the exceptions go uncaught al the way to the main()

src/bus_monitor.cpp
```diff
@@ -10,6 +10,77 @@
 namespace panel
 {
 
+static void resetLEDState()
+{
+    std::cout << "Reset LED state" << std::endl;
+    std::vector<std::string> ledsPathOnBasePanel{
```
santoshpuranik, line 16: Please check if this can be constexpr. We don't want to create the vector every time in the function

src/bus_monitor.cpp
```diff
@@ -10,6 +10,77 @@
...
+        else
+        {
+            std::cerr
+                << "Failed to read asserted property. Unable to reset led "
+                << led << std::endl;
+        }
+    }
+}
+
+void PanelPresence::readEverBasePresentProperty(
```
santoshpuranik, line 53: Please consider renaming this function (remove the "Ever" from the name).

src/panel_app_main.cpp
```diff
@@ -128,14 +128,31 @@ int main(int, char**)
...
                 std::get<0>((baseDataMap.find(imValue))->second),
                 std::get<1>((baseDataMap.find(imValue))->second),
                 panel::types::PanelType::BASE);
+
+            auto& baseObjPath =
+                std::get<2>((baseDataMap.find(imValue))->second);
+
+            // if it is base panel of Everest, register for its presence to
+            // enable CM.
+            if (sdbusplus::message::object_path(baseObjPath).filename() ==
```
santoshpuranik, line 147: Why check just for panel0 here? We could just check for constants::everBaseDbusObj, right?

santoshpuranik (changes requested):

src/bus_monitor.cpp
SunnySrivastava1984: This function will be called only when Everest base panel will se set from false to true i.e. only at panel CM.
How frequently this will be called? Just trying to understand if it will have much effect?

src/bus_monitor.cpp
SunnySrivastava1984: Fixed.

src/bus_monitor.cpp
SunnySrivastava1984: Done.

src/panel_app_main.cpp
SunnySrivastava1984: Yes we can. I thought of using the bmcweb way. I will change that.

PR #157, opened 2022-08-10
Panel CM in Everest

SunnySrivastava1984:
The commit make changes related to panel CM in case of
Everest system.
As a process of which, in case, base panel on Everest
undergoes CM, state of all the LEDs on base panel needs to
be restored.

### DIFF
--- a/include/bus_monitor.hpp
+++ b/include/bus_monitor.hpp
@@ -52,6 +52,14 @@ class PanelPresence
      * @param[in] msg - pointer to the msg sent by the PropertiesChanged signal.
      */
     void readPresentProperty(sdbusplus::message::message& msg);
+
+    /**
+     * @brief Base panel "Present" property callback.
+     * It resets LEDs state on base panel and sets the transport key.
+     *
+     * @param[in] msg - pointer to the msg sent by the PropertiesChanged signal.
+     */
+    void readBasePresentProperty(sdbusplus::message::message& msg);
 };
 
 /** @class PELListener
--- a/src/bus_monitor.cpp
+++ b/src/bus_monitor.cpp
@@ -10,6 +10,81 @@
 namespace panel
 {
 
+static void resetLEDState()
+{
+    static constexpr auto ledsPathOnBasePanel = {
+        "/xyz/openbmc_project/led/groups/bmc_booted",
+        "/xyz/openbmc_project/led/groups/power_on",
+        "/xyz/openbmc_project/led/groups/enclosure_identify",
+        "/xyz/openbmc_project/led/groups/platform_system_attention_indicator",
+        "/xyz/openbmc_project/led/groups/partition_system_attention_indicator",
+        "/xyz/openbmc_project/led/groups/enclosure_fault"};
+
+    for (const auto& led : ledsPathOnBasePanel)
+    {
+        auto res = utils::readBusProperty<std::variant<bool>>(
+            "xyz.openbmc_project.LED.GroupManager", led,
+            "xyz.openbmc_project.Led.Group", "Asserted");
+
+        if (auto asserted = std::get_if<bool>(&res))
+        {
+            if (*asserted)
+            {
+                try
+                {
+                    utils::writeBusProperty<bool>(
+                        "xyz.openbmc_project.LED.GroupManager", led,
+                        "xyz.openbmc_project.Led.Group", "Asserted", false);
+
+                    utils::writeBusProperty<bool>(
+                        "xyz.openbmc_project.LED.GroupManager", led,
+                        "xyz.openbmc_project.Led.Group", "Asserted", true);
+                }
+                catch (const sdbusplus::exception::SdBusError& e)
+                {
+                    std::cerr << "Write call for led " << led
+                              << " failed with exception " << e.what()
+                              << std::endl;
+                }
+            }
+        }
+        else
+        {
+            std::cerr
+                << "Failed to read asserted property. Unable to reset led "
+                << led << std::endl;
+        }
+    }
+}
+
+void PanelPresence::readBasePresentProperty(sdbusplus::message::message& msg)
+{
+    if (msg.is_method_error())
+    {
+        std::cerr << "\n Error in reading base panel presence signal "
+                  << std::endl;
+    }
+    std::string interface;
+    types::PropertyValueMap propMap;
+    msg.read(interface, propMap);
+    const auto itr = propMap.find("Present");
+    if (itr != propMap.end())
+    {
+        if (auto present = std::get_if<bool>(&(itr->second)))
+        {
+            if (*present)
+            {
+                resetLEDState();
+            }
+        }
+        else
+        {
+            std::cerr << "\n Error reading base panel present property"
+                      << std::endl;
+        }
+    }
+}
+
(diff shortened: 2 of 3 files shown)


src/bus_monitor.cpp
```diff
@@ -10,6 +10,84 @@
...
+            "xyz.openbmc_project.Led.Group", "Asserted");
+
+        if (auto asserted = std::get_if<bool>(&res))
+        {
+            if (*asserted)
+            {
+                try
+                {
+                    std::cout << "Led asserted = " << led << std::endl;
+                    utils::writeBusProperty<bool>(
```
SunnySrivastava1984, line 37: > Santosh: These can throw - and I don't think we should let the exceptions go uncaught al the way to the main()

> Sunny: FIxed

src/bus_monitor.cpp
```diff
@@ -10,6 +10,84 @@
 namespace panel
 {
 
+static void resetLEDState()
+{
+    std::cout << "Reset LED state" << std::endl;
+    std::vector<std::string> ledsPathOnBasePanel{
```
SunnySrivastava1984, line 16: > Santosh: Please check if this can be constexpr. We don't want to create the vector every time in the function

> Sunny: This function will be called only when present property of Everest base panel will be set from false to true i.e. only at panel CM and at the time of system coming up.
How frequently this will be called? Just trying to understand if it will have much effect?

src/bus_monitor.cpp
```diff
@@ -10,6 +10,84 @@
...
+        else
+        {
+            std::cerr
+                << "Failed to read asserted property. Unable to reset led "
+                << led << std::endl;
+        }
+    }
+}
+
+void PanelPresence::readBasePresentProperty(sdbusplus::message::message& msg)
```
SunnySrivastava1984, line 62: > Santosh: Please consider renaming this function (remove the "Ever" from the name).

>Sunny: Done

src/panel_app_main.cpp
```diff
@@ -83,14 +83,30 @@ int main(int, char**)
...
                 std::get<0>((baseDataMap.find(imValue))->second),
                 std::get<1>((baseDataMap.find(imValue))->second),
                 panel::types::PanelType::BASE);
+
+            auto& baseObjPath =
+                std::get<2>((baseDataMap.find(imValue))->second);
+
+            // if it is base panel of Everest, register for its presence to
+            // enable CM.
+            if (baseObjPath == panel::constants::everBaseDbusObj)
```
SunnySrivastava1984, line 102: > Santosh: Why check just for panel0 here? We could just check for constants::everBaseDbusObj, right?

> Sunny: Yes we can. I thought of using the bmcweb way. I will change that.

src/bus_monitor.cpp
santoshpuranik: Just good coding practice?

src/bus_monitor.cpp
```diff
@@ -10,6 +10,84 @@
...
+{
+    if (msg.is_method_error())
+    {
+        std::cerr << "\n Error in reading panel presence signal " << std::endl;
```
santoshpuranik, line 66: error in reading base panel presence ...

src/bus_monitor.cpp
santoshpuranik: BTW, don't take the constexpr literally, you cannot create a local std::vector that is a constexpr. You can however, do something like a std::array<const char*>

src/bus_monitor.cpp
```diff
@@ -10,6 +10,84 @@
...
+            "xyz.openbmc_project.LED.GroupManager", led,
+            "xyz.openbmc_project.Led.Group", "Asserted");
+
+        if (auto asserted = std::get_if<bool>(&res))
+        {
+            if (*asserted)
+            {
+                try
+                {
+                    std::cout << "Led asserted = " << led << std::endl;
```
santoshpuranik, line 36: Nit: The cout can be outside the try

src/bus_monitor.cpp
SunnySrivastava1984: Yup.

include/bus_monitor.hpp
```diff
@@ -52,6 +52,14 @@ class PanelPresence
      * @param[in] msg - pointer to the msg sent by the PropertiesChanged signal.
      */
     void readPresentProperty(sdbusplus::message::message& msg);
+
+    /**
+     * @brief Everest base panel "Present" property callback.
```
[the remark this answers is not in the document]
SunnySrivastava1984: Done

src/bus_monitor.cpp
SunnySrivastava1984: Done

src/bus_monitor.cpp
```diff
@@ -10,6 +10,85 @@
...
+    }
+    std::string interface;
+    types::PropertyValueMap propMap;
+    msg.read(interface, propMap);
+    const auto itr = propMap.find("Present");
+    if (itr != propMap.end())
+    {
+        if (auto present = std::get_if<bool>(&(itr->second)))
+        {
+            transport->setTransportKey(*present);
```
SunnySrivastava1984, line 77: Dont set transport key of basae panel as base panel chip resides on bmc card and changing transport key will trigger code update flow.

src/bus_monitor.cpp
```diff
@@ -10,6 +10,85 @@
 namespace panel
 {
 
+static void resetLEDState()
+{
+    std::cout << "Reset LED state" << std::endl;
+    std::array<const char*, 6> ledsPathOnBasePanel{
```
santoshpuranik, line 16: Please see if this can be a static constexpr

src/bus_monitor.cpp
santoshpuranik: You still have to handle this

santoshpuranik (changes requested):
Looks good, a few comments noted down earlier need to be handled

src/bus_monitor.cpp
SunnySrivastava1984: Yes

src/bus_monitor.cpp
```diff
@@ -10,6 +10,83 @@
...
+    {
+        auto res = utils::readBusProperty<std::variant<bool>>(
+            "xyz.openbmc_project.LED.GroupManager", led,
+            "xyz.openbmc_project.Led.Group", "Asserted");
+
+        if (auto asserted = std::get_if<bool>(&res))
+        {
+            if (*asserted)
+            {
+                std::cout << "Led asserted = " << led << std::endl;
```
jinuthomas, line 34: this print is not needed

src/bus_monitor.cpp
```diff
@@ -10,6 +10,83 @@
 namespace panel
 {
 
+static void resetLEDState()
+{
+    std::cout << "Reset LED state" << std::endl;
```
jinuthomas, line 15: would avoid this print , but liveable

jinuthomas (approved):
just remove prints

src/bus_monitor.cpp
SunnySrivastava1984: Removed.

PR #159, opened 2022-08-31
Enable code update feature for op-panel

PriyangaRamasamy:
Enabling code update feature for both base and lcd
panel firmware. The code update takes place once the
panel's physical presence is detected.

If the panel firmware code is latest then code update
is not required, else ibm-panel app does panel firmware
code update to the latest version available.

To do code update panel firmware needs to support a minimum
version level.

Test:
Tested for the below cases and code works as expected.

Case 1: Successful code update
Case 2: Panel firmware is latest and code update not required
Case 3: Abort code update and log PEL as the panel firmware doesn't
        support minimum version level.

Change-Id: Ic72d70c962f7d64fd38839f67a804e9bd34f8fc4


jinuthomas (changes requested):
Not required for 1020.10

PR #165, opened 2022-11-29
Handle PEL deletion and update function 64 flow

SunnySrivastava1984:
The commit implement changes to handle PEL deletion event, where in case last stored PEL was deleted then function 11 to 19 will be set to disabled mode.

It also make changes to the flow of function 64, where only after execution of function 64 the code will fetch and filter the list of diagnostic SRCs required to be displayed via this function.

Test Pending...

### DIFF
--- a/include/bus_monitor.hpp
+++ b/include/bus_monitor.hpp
@@ -96,13 +96,16 @@ class PELListener
     }
 
     /**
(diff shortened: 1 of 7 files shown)


SunnySrivastava1984:
Pushed via different commit

PR #176, opened 2023-01-25
Update function state and data on PEL deletion

SunnySrivastava1984:
The commit fixes issue where function 11 to 13 were not getting enabled after a cycle where a PEL was logged then deleted and then logged.

It also fixes the issue where function 11 was holding old SRC data once the above cycle was performed.


SunnySrivastava1984:
Fixes STG Defect 395986

jinuthomas (approved):

PR #180, opened 2023-02-02
Op-Panel changes for new system Bonnell

GiridhariKrishna:
Addition of LCD Device Path, IM Keyword for New System Bonnell support.

### DIFF
--- a/include/const.hpp
+++ b/include/const.hpp
@@ -8,9 +8,11 @@ namespace constants
 {
 
 static constexpr auto baseDevPath = "/dev/i2c-3";
+static constexpr auto bonnellBaseDevPath = "/dev/i2c-2";
 static constexpr auto rainLcdDevPath = "/dev/i2c-7";
 static constexpr auto everLcdDevPath = "/dev/i2c-28";
 static constexpr auto tacomaLcdDevPath = "/dev/i2c-0";
+static constexpr auto bonnellLcdDevPath = rainLcdDevPath;
 
 static constexpr auto devAddr = 0x5a;
 
@@ -22,6 +24,7 @@ static constexpr auto rainBaseDbusObj =
 static constexpr auto everBaseDbusObj =
     "/xyz/openbmc_project/inventory/system/chassis/motherboard/dasd_backplane/"
     "panel0";
+static constexpr auto bonnellBaseDbusObj = rainBaseDbusObj;
 static constexpr auto rainLcdDbusObj = "/xyz/openbmc_project/inventory/system/"
                                        "chassis/motherboard/lcd_op_panel_hill";
 static constexpr auto everLcdDbusObj =
@@ -29,11 +32,13 @@ static constexpr auto everLcdDbusObj =
     "panel1";
 static constexpr auto everBMCObj =
     "/xyz/openbmc_project/inventory/system/chassis/motherboard/bmc";
+static constexpr auto bonnellLcdDbusObj = rainLcdDbusObj;
 
 static constexpr auto rain2s2uIM = "50001001";
 static constexpr auto rain2s4uIM = "50001000";
 static constexpr auto rain1s4uIM = "50001002";
 static constexpr auto everestIM = "50003000";
+static constexpr auto bonnellIM = "50004000";
 
 static constexpr auto imInterface = "com.ibm.ipzvpd.VSBP";
 static constexpr auto imKeyword = "IM";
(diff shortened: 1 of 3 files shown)


geissonator:
add to approvelist

src/panel_app_main.cpp
```diff
@@ -21,13 +21,17 @@ panel::types::PanelDataMap baseDataMap = {
       panel::constants::rainBaseDbusObj}},
     {panel::constants::everestIM,
      {panel::constants::baseDevPath, panel::constants::devAddr,
-      panel::constants::everBaseDbusObj}}};
+      panel::constants::everBaseDbusObj}},
+    {panel::constants::bonnellIM,
+     {panel::constants::baseDevPath, panel::constants::devAddr,
+      panel::constants::bonelBootFailChipFRU}}};
```
jinuthomas, line 27: bonelBootFailChipFRU why name change should it not be bonnellBaseDbusObj

jinuthomas (changes requested):
I also see Bonnell missing in static const types::PICFRUPathMap bootFailPIC 

I HAVE NOT COMPLETED THE REVIEW, could you please take a relook at if any changes needed elsewhere or not .

include/const.hpp
```diff
@@ -29,11 +30,14 @@ static constexpr auto everLcdDbusObj =
     "panel1";
 static constexpr auto everBMCObj =
     "/xyz/openbmc_project/inventory/system/chassis/motherboard/bmc";
+static constexpr auto bonelBootFailChipFRU = systemDbusObj;
```
PriyangaRamasamy, line 33: bonnell* spellcheck

PriyangaRamasamy: @GiridhariKrishna  As we discussed today, bonnellBootFailChipFRU is systemDbusObj thats right, but the place where you have fed this information is wrong. The boot fail PIC FRU details are in this map https://github.com/ibm-openbmc/ibm-panel/blob/1050/include/const.hpp#L80 . please update this map for bonnell.

and in this baseDataMap https://github.com/ibm-openbmc/ibm-panel/blob/1050/src/panel_app_main.cpp#L12
you need to provide bonnellBaseDbusObj path and not the bootFailPIC fru path.

src/panel_app_main.cpp
```diff
@@ -21,13 +21,17 @@ panel::types::PanelDataMap baseDataMap = {
...
 
 std::string getInputDevicePath(const std::string& imValue)
 {
     if (imValue == panel::constants::rain2s2uIM ||
         imValue == panel::constants::rain2s4uIM ||
-        imValue == panel::constants::rain1s4uIM)
+        imValue == panel::constants::rain1s4uIM ||
+        imValue == panel::constants::bonnellIM)
     {
         return "/dev/input/by-path/platform-1e78a400.i2c-bus-event-joystick";
```
PriyangaRamasamy, line 36: can you please point me to the place where you confirmed that this is the input device path for bonnell panel.

include/const.hpp
GiridhariKrishnan: > bonnell* spellcheck

It's intentional

src/panel_app_main.cpp
jinuthomas: This will be the same as in Rainier as the i2c for LCD panel is same.

jinuthomas (changes requested):
The configurations are wrong

include/const.hpp
PriyangaRamasamy: that should be ok to give complete word - bonnell with right spelling. its your call.

PriyangaRamasamy: > and in this baseDataMap https://github.com/ibm-openbmc/ibm-panel/blob/1050/src/panel_app_main.cpp#L12
> you need to provide bonnellBaseDbusObj path and not the bootFailPIC fru path.

can you please address this comment?

src/panel_app_main.cpp
PriyangaRamasamy: ok

include/const.hpp
GiridhariKrishnan: Yes, initially thought so but confused with systemDbusobj, will change that.

GiridhariKrishnan: Sorry, changed that already, will update.

jinuthomas (changes requested):
overall looks good , once the nit is done , we can merge this

jinuthomas (approved):

PR #182, opened 2023-06-02
Display SRC word 0 for terminating SRCs and consider a checkstop a terminating SRC

spinler:
Bit 0 in hex word 5 of BMC SRCs indicates the PEL is for a checkstop, which should be posted on the panel so that the op panel isn't just left with whatever was posted previous to the hostboot/OS crash.

Also, display word 0 of the SRC event ID field, which is the ascii string field of the SRC, like the BD50E510.

### DIFF
--- a/src/bus_monitor.cpp
+++ b/src/bus_monitor.cpp
@@ -224,7 +224,7 @@ void PELListener::PELEventCallBack(sdbusplus::message::message& msg)
                             // code is for BMC i.e "BD". Send it
                             // directly to display.
                             utils::sendCurrDisplayToPanel(
-                                hexWords.at(4), std::string{}, transport);
+                                hexWords.at(0), std::string{}, transport);
                         }
                         executor->storeLastPelEventId(*eventId);
                         lastPelObjPath = objPath;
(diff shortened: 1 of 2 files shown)


src/bus_monitor.cpp
```diff
@@ -217,7 +217,7 @@ void PELListener::PELEventCallBack(sdbusplus::message::message& msg)
                         types::Byte byte =
                             ::strtoul(valueAtIndexZero, nullptr, 16);
 
-                        if ((byte & constants::terminatingBit) != 0x00 &&
+                        if ((byte & constants::terminatingBits) != 0x00 &&
```
SunnySrivastava1984, line 220: As per the spec, 0th bit and 2nd bit from MSB end will account for check stop and terminating bit respectively.
Performing an AND with 0xC in this case will not detect terminating bit.

src/bus_monitor.cpp
```diff
@@ -217,14 +217,14 @@ void PELListener::PELEventCallBack(sdbusplus::message::message& msg)
...
```
SunnySrivastava1984, line 220: As per the spec, 0th bit and 2nd bit from MSB end will account for check stop and terminating bit respectively.
Performing an AND with 0xC in this case will not detect terminating bit.

SunnySrivastava1984 (changes requested):

src/bus_monitor.cpp
spinler: Fixed.

SunnySrivastava1984 (approved):

jinuthomas (approved):

PR #188, opened 2023-07-10
Set operating mode based on BIOS table

SunnySrivastava1984:
Unlike GUI, Op-panel sets system operating mode based on value of "PowerRestorePolicy" and "RebootPolicy".
To keep the funcytionality in sync, panel code has been modified to update panel state only in case BIOS table has been modified.

Test:
Change system operating mode from GUI, same should be reflected in Panel and vice versa.

### DIFF
--- a/include/bus_monitor.hpp
+++ b/include/bus_monitor.hpp
@@ -237,10 +237,10 @@ class SystemStatus
     void listenBootProgressState();
 
     /**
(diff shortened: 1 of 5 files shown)


SunnySrivastava1984:
jenkins -> run tests please

SunnySrivastava1984:
jenkins -> run tests please

PR #203, opened 2023-10-06
Creates an empty logger library

RekhaAparna01:
Creates an empty logger library
Change-Id: I26288f8a00415b519aa9d8c0304e8f038e0febcd

### DIFF
--- a/logger/include/main.hpp
+++ b/logger/include/main.hpp
@@ -1,0 +1,13 @@
+#pragma once
+
+#include <iostream>
+#include <source_location>
+
+/*
+ * @brief Define API's to append log messages to a buffer
+ */
+namespace logging
+{
+void logMessage(std::string msg, const std::source_location& location);
+
+} // namespace logging
--- a/logger/meson.build
+++ b/logger/meson.build
@@ -1,0 +1,3 @@
+logger_a = static_library('logger', 'src/main.cpp', include_directories
+                          : 'include', install
+                          : true, )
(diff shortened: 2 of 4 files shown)


geissonator:
add to approvelist

logger/meson.build
```diff
@@ -0,0 +1,3 @@
+logger_a = static_library('logger', 'src/main.cpp', include_directories
```
Alpana07, line 1: I think we can do this in same existing meson.build.
until the team has some intention to keep it separate .

logger/include/main.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <iostream>
+#include <source_location>
+
+/*
+ * @brief Define API's to append log messages to a buffer
```
Alpana07, line 7: add doxygen just before the function declaration. and keep namespace in outer most.

Alpana07: Maintain this formate for doxygen-

* @brief This API does ........in 1 line
*  more description if needed
* 
* @ param[in] message -- tell its purpose 
*  @ param[in] locationDetails -- tell its purpose 
*  @ param[out] out -- in case you have to share some output to called function.
*
* @return -- what It returns in good case/ bad case , if void keep it blank.


Something can be changed also based on need, but in general we use this format.

Alpana07 (commented):
Please write some details in commit message , as what this commit is going to enable.
so this the format you can follow-

Title - Creates an empty logger library (max 50 character, I guess )

Description (max 80 characters in 1 line) -
This commit enables.....
OR it will fix this current issue....talk about current problem

Test:
You can talk about how did you verify the patch, and some traces/snippet of test result.

PR #206, opened 2023-12-01
API to fetch logs from buffer

RekhaAparna01:
This commit implements an API to fetch specified
number of logs from the buffer and dump them to the file

Test:
Tested by adding traces for library

Change-Id: I09c2126c8caefe1e804304f1697288d8a077b168

### DIFF
--- a/logger/include/logger.hpp
+++ b/logger/include/logger.hpp
@@ -45,4 +45,14 @@ void logMessage(
     const std::string& message,
     const std::source_location& location = std::source_location::current());
 
+/**
+ * @brief Fetches specified number of logs and dump to a file
+ *
+ * The API fetches the specified number of logs from the buffer
+ * and dumps the logs into a file.
+ *
+ * @param[in] noOfLogs - Number of logs to be fetched
+ */
+void fetchNLogs(size_t noOfLogs);
+
 }; // namespace Logger
--- a/logger/src/logger.cpp
+++ b/logger/src/logger.cpp
@@ -1,6 +1,7 @@
 #include "logger.hpp"
 
 #include <ctime>
+#include <fstream>
 
 namespace Logger
 {
@@ -9,6 +10,7 @@ static std::map<Loglevel, std::string> logLevelMap = {
     {Loglevel::INFO, "INFO"}, {Loglevel::CRITICAL, "CRITICAL"}};
 
 size_t byteCount = 0;
+const std::string logFilePath = "/var/lib/vpd/panellog.log";
 
 /**
  * @brief Generates a timestamp string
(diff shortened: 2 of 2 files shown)


logger/include/logger.hpp
```diff
@@ -45,4 +45,14 @@ void logMessage(
     const std::string& message,
     const std::source_location& location = std::source_location::current());
 
+/**
+ * @brief Fetches the latest logs from the buffer
```
SunnySrivastava1984, line 49: Re-phrase? It fetches number of loges as specified in the parameter.  re-word accordingly.

logger/include/logger.hpp
```diff
@@ -45,4 +45,14 @@ void logMessage(
...
+ *
+ * The API fetches latest N logs from the buffer and
+ * adds the logs in a file.
+ *
+ * @param[in] nooflogs - Number of logs to be fetched
```
SunnySrivastava1984, line 54: Camel case: - noOfLogs?

logger/src/logger.cpp
```diff
@@ -56,4 +57,20 @@ void logMessage(const std::string& message,
     logMessage(Loglevel::INFO, message, location);
 }
 
+void fetchNlogs(size_t noofLogs)
+{
+    std::fstream logfile("/tmp/logdata.txt", std::ios::trunc | std::ios::out);
```
SunnySrivastava1984, line 62: Proper error handling is missing. Please fix that.

logger/src/logger.cpp
```diff
@@ -56,4 +57,20 @@ void logMessage(const std::string& message,
...
+    if (noofLogs > circularBuffer.size())
+    {
+        noofLogs = circularBuffer.size();
+    }
+    auto iterator = circularBuffer.begin();
+    std::advance(iterator, circularBuffer.size() - noofLogs);
+    while (iterator != circularBuffer.end())
```
SunnySrivastava1984, line 69: Add empty space in between two blocks of code.

SunnySrivastava1984 (changes requested):
Have a proper commit message please. The commit not only fetches but dumps into a file as well.

logger/src/logger.cpp
RekhaAparna01: I have not got any error while opening file. The file is getting created if the file is not present.

logger/src/logger.cpp
RekhaAparna01: done

logger/include/logger.hpp
RekhaAparna01: done

logger/include/logger.hpp
RekhaAparna01: done

logger/src/logger.cpp
SunnySrivastava1984: Error handling is required for negative case. Please handle the case for any failure with file handling.

logger/src/logger.cpp
```diff
@@ -56,4 +57,31 @@ void logMessage(const std::string& message,
     logMessage(Loglevel::INFO, message, location);
 }
 
+void fetchNLogs(size_t noOfLogs)
+{
+    try
+    {
+        std::fstream logfile("/tmp/logdata.txt",
```
SunnySrivastava1984, line 64: add error handling for stream object.

SunnySrivastava1984 (changes requested):

logger/src/logger.cpp
```diff
@@ -56,4 +57,36 @@ void logMessage(const std::string& message,
...
```
branupama, line 64: can you we keep this log file path in header as a const variable.
and usually log files will end with .log extension.

logger/src/logger.cpp
```diff
@@ -56,4 +57,36 @@ void logMessage(const std::string& message,
...
+                             std::ios::trunc | std::ios::out);
+        if (!logfile.is_open())
+        {
+            throw std::runtime_error("Error opening the file!");
```
branupama, line 68: append file name as well, so will come to know which file failed to open ..

logger/src/logger.cpp
SunnySrivastava1984: Constant is ok, but I don't agree with moving it to the header file.
Since the use is local to the file, declare in the scope of this file itself.

RekhaAparna01: done

logger/src/logger.cpp
RekhaAparna01: done

SunnySrivastava1984 (approved):

PR #211, opened 2024-07-09
API to reboot BMC

RekhaAparna01:
This commit implements function 75 an API to trigger BMC reboot. Function 75 is enabled only when in manual operating mode and CE mode is enabled.

output:
A1003075

Test:
1. After initiating BMC reboot from panel, observed the following logs from the journal
phosphor-bmc-state-manager[659]: Setting the RequestedBMCTransition field to xyz.openbmc_project.State.BMC.Transition.Reboot
phosphor-bmc-state-manager[659]: Setting the BMCState field to xyz.openbmc_project.State.BMC.BMCState.NotReady

2. Tested by killing the bmcweb process.
Killed the bmcweb processs. Observed the GUI is frozen and then initiated reboot from panel.
After reboot the GUI is back and able to login.

Change-Id: I9f9a1482fbff8cf3ade6c98b6dcdc57bd62b6d36

### DIFF
--- a/src/executor.cpp
+++ b/src/executor.cpp
@@ -161,6 +161,10 @@ void Executor::executeFunction(const types::FunctionNumber funcNumber,
                 execute74();
                 break;
 
+            case 75:
+                execute75();
+                break;
+
             default:
                 break;
         }
@@ -1044,6 +1048,16 @@ void Executor::execute74()
         }
     }
     displayExecutionStatus(74, std::vector<types::FunctionNumber>(), true);
+}
+
+void Executor::execute75()
+{
+    utils::writeBusProperty<std::string>(
+        "xyz.openbmc_project.State.BMC", "/xyz/openbmc_project/state/bmc0",
+        "xyz.openbmc_project.State.BMC", "RequestedBMCTransition",
+        "xyz.openbmc_project.State.BMC.Transition.Reboot");
+
+    utils::sendCurrDisplayToPanel("REBOOTING BMC", " ", transport);
 }
 
 types::ReturnStatus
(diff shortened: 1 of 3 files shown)


SunnySrivastava1984 (approved):
Please update the latest test case as well in the commit message.

src/executor.cpp
```diff
@@ -160,6 +160,9 @@ void Executor::executeFunction(const types::FunctionNumber funcNumber,
             case 74:
                 execute74();
                 break;
+            case 75:
```
jinuthomas, line 163: please add a new line above the case

src/executor.cpp
```diff
@@ -1046,6 +1049,14 @@ void Executor::execute74()
     displayExecutionStatus(74, std::vector<types::FunctionNumber>(), true);
 }
 
+void Executor::execute75()
+{
+    utils::writeBusProperty<std::string>(
+        "xyz.openbmc_project.State.BMC", "/xyz/openbmc_project/state/bmc0",
+        "xyz.openbmc_project.State.BMC", "RequestedBMCTransition",
+        "xyz.openbmc_project.State.BMC.Transition.Reboot");
+}
```
jinuthomas, line 1058: add the below line  as last line in your function: 
utils::sendCurrDisplayToPanel("REBOOTING  BMC", "              ", transport);

src/executor.cpp
RekhaAparna01: Done

src/executor.cpp
RekhaAparna01: Updated

jinuthomas (approved):

PR #214, opened 2024-07-12
1050: API to reboot BMC

RekhaAparna01:
This commit implements function 75 an API to trigger BMC reboot. Function 75 is enabled only when in manual operating mode and CE mode is enabled.

Test:
1. After initiating BMC reboot from panel, observed the following logs from the journal phosphor-bmc-state-manager[659]: Setting the RequestedBMCTransition field to xyz.openbmc_project.State.BMC.Transition.Reboot phosphor-bmc-state-manager[659]: Setting the BMCState field to xyz.openbmc_project.State.BMC.BMCState.NotReady

2. Tested by killing the bmcweb process. Killed the bmcweb processs. Observed the GUI is frozen and then initiated reboot from panel. After reboot the GUI is back and able to login.

Change-Id: I184f47e97241335417d239d10ba395206cdf51a9

### DIFF
--- a/include/executor.hpp
+++ b/include/executor.hpp
@@ -259,6 +259,11 @@ class Executor
     void execute74();
 
     /**
+     * @brief An api to execute function 75.
+     */
+    void execute75();
+
+    /**
      * @brief Api to get PEL eventId.
      *
      * This is a helper function to get the eventId(SRC) data for the PEL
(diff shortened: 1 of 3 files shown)


RekhaAparna01:
jenkins run tests please

SunnySrivastava1984 (approved):
Cherry-picked from 1060.

SunnySrivastava1984 (commented):
This needs to be merged only after 1050.20 is done.

jinuthomas (approved):

PR #223, opened 2025-01-28
Panel migrates to use libpldm APIs

PriyangaRamasamy:


src/pldm_fw.cpp
```diff
@@ -105,10 +115,50 @@ int PldmFramework::openMctpDemuxTransport()
...
+            << "openAfMctpTransport: Failed to init AF MCTP transport. rc = "
+            << rc << std::endl;
+        throw FunctionFailure("Failed to init AF MCTP transport");
+    }
+
+    rc = pldm_transport_af_mctp_map_tid(impl.afMctp, tid, tid);
+    if (rc)
+    {
+        std::cerr
+            << "openAfMctpTransport: Failed to setup tid to eid mapping. rc = "
```
branupama: tid and eid should be capital.

branupama (changes requested):

src/pldm_fw.cpp
```diff
@@ -7,33 +7,159 @@
...
+        throw FunctionFailure("Failed to get pollfd");
+    }
+    return pollfd.fd;
+}
+
+[[maybe_unused]] int PldmFramework::openAfMctpTransport()
+{
+    impl.afMctp = nullptr;
+    int rc = pldm_transport_af_mctp_init(&impl.afMctp);
+    if (rc)
```
branupama: Better to use (rc != 0) as rc is int type, will gives more clarity to reader.

same for below as well.

src/pldm_fw.cpp
```diff
@@ -59,7 +59,16 @@ int PldmFramework::openPldm()
         return fd;
     }
 
+#if defined(PLDM_TRANSPORT_WITH_MCTP_DEMUX)
     fd = openMctpDemuxTransport();
+#elif defined(PLDM_TRANSPORT_WITH_AF_MCTP)
+    fd = openAfMctpTransport();
+#else
+    std::cerr << "open: No valid transport defined!" << std::endl;
```
branupama: No valid PLDM transport ?

src/pldm_fw.cpp
```diff
@@ -59,7 +59,16 @@ int PldmFramework::openPldm()
...
+    throw FunctionFailure("Required host dump action via PLDM is not allowed: "
```
branupama: does openPldm() caller handling exception ?
do we need to send -1 in case of error ? to create PEL in the caller .?

src/pldm_fw.cpp
```diff
@@ -105,10 +115,50 @@ int PldmFramework::openMctpDemuxTransport()
...
+[[maybe_unused]] int PldmFramework::openAfMctpTransport()
+{
+    impl.afMctp = nullptr;
+    int rc = pldm_transport_af_mctp_init(&impl.afMctp);
+    if (rc)
+    {
+        std::cerr
+            << "openAfMctpTransport: Failed to init AF MCTP transport. rc = "
+            << rc << std::endl;
+        throw FunctionFailure("Failed to init AF MCTP transport");
```
branupama: does openPldm() caller handling exception ?
do we need to send -1 in case of error ? to create PEL in the caller .?

same for below as well

branupama (changes requested):

src/pldm_fw.cpp
PriyangaRamasamy: this method returns file descriptor and not rc. we cannot return rc to the caller. will add a catch block in caller  and log PEL there

src/pldm_fw.cpp
PriyangaRamasamy: I don't find a value in explicitly checking for 0. 
reader should know the return type of pldm_transport_af_mctp_init to get the clarity.

src/pldm_fw.cpp
PriyangaRamasamy: there is nothing called as PLDM TRANSPORT.

branupama:
> This change migrates the application off the deprecated "requester" APIs in
libpldm.
I dint understand this statement

include/pldm_fw.hpp
```diff
@@ -21,12 +24,26 @@ class PldmFramework
...
      * @brief Destructor
      */
-    ~PldmFramework() = default;
+    ~PldmFramework()
+    {
+        int rc = pldm_instance_db_destroy(pldmInstanceIdDb);
+        if (rc)
+        {
+            std::cout << "pldm_instance_db_destroy failed, rc =" << rc << "\n";
+        }
```
branupama: should we set pldmInstanceIdDb to nullptr here ?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,109 @@
...
-            "xyz.openbmc_project.PLDM.Requester", "GetInstanceId");
-        method.append(mctpEid);
-        auto reply = bus.call(method);
-        reply.read(instanceId);
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        rc = pldm_instance_id_alloc(pldmInstanceIdDb, tid, &instanceID);
     }
-    catch (const sdbusplus::exception::SdBusError& e)
+
+    if (rc)
```
branupama: should it be check on -EAGAIN ?

branupama (changes requested):

src/pldm_fw.cpp
PriyangaRamasamy: if (rc == -EAGAIN) this is already checked at line 23 - that is to try to allocate instance id again..

when the first time we hit -EAGAIN - we attempt to retry.. but when the second time we hit -EAGAIN or any rc, we throw the error..

include/pldm_fw.hpp
PriyangaRamasamy: yes. will do that

include/pldm_fw.hpp
```diff
@@ -73,7 +74,13 @@ class PldmFramework
     /** pldm transport instance  */
     pldm_transport* pldmTransport = nullptr;
 
-    pldm_transport_mctp_demux* mctpDemux = nullptr;
+    union TransportImpl
```
souvik1914581: Any reason why we are using `union` here instead of `std::variant<>` ?

jinuthomas (changes requested):
At a high level see two changes I feel not apt 
1. the return code checks , no where it tells me that a positive value is a failure.
2. the multi level throw in code , is that even required , does it help us in any way

include/pldm_fw.hpp
PriyangaRamasamy: std::variant is for runtime detection of the types. here we will decide which transport layer has to be used at compile time itself. so we use union to allocate the shared memory space for both the types.. Based on the meson option provided at compile time, we will decide the transport type.

include/pldm_fw.hpp
```diff
@@ -21,12 +25,30 @@ class PldmFramework
...
     /**
      * @brief Destructor
      */
-    ~PldmFramework() = default;
+    ~PldmFramework()
+    {
+        int rc = pldm_instance_db_destroy(pldmInstanceIdDb);
+        if (rc)
+        {
+            if (pldmInstanceIdDb)
```
branupama: we don't need this check here and we can directly assign nullptr to pldmInstanceIdDb, as we going out of scope.

src/pldm_fw.cpp
```diff
@@ -7,33 +7,162 @@
...
     {
-        std::cerr << e.what() << std::endl;
+        std::cerr << "Return code = " << rc << std::endl;
```
RekhaAparna01: Can the error statement be: "Failed to get instance id. Return code = ". This will help to know which function failed?

PriyangaRamasamy:
> At a high level see two changes I feel not apt
> 
> 1. the return code checks , no where it tells me that a positive value is a failure.
> 2. the multi level throw in code , is that even required , does it help us in any way

1. These are the return codes from **libpldm APIs** which returns errno on failure and 0 on success. 
Eg: https://github.com/openbmc/libpldm/blob/8fd26555be33fadb0c0345897ddcbec859fda1b5/src/requester/instance-id.c#L4.
2. As a caller PldmFramework::sendPanelFunctionToPhyp() -  this API will know for which function number the execution failed and it will log a PEL for the specific reason which is thrown from the point of failure. Yes it helps us in logging PEL at the caller place.

src/pldm_fw.cpp
branupama: okay

src/pldm_fw.cpp
branupama: is this is pending ?

branupama (changes requested):

src/pldm_fw.cpp
```diff
@@ -7,33 +7,162 @@
...
+
+int PldmFramework::openPldm()
+{
+    auto fd = -1;
+    if (pldmTransport)
+    {
+        std::cerr << "open: pldmTransport already setup!" << std::endl;
+        throw FunctionFailure("Required host dump action via PLDM is not "
+                              "allowed due to openPldm failed");
+        return fd;
```
RekhaAparna01: since you are throwing exception, return is not required.

include/pldm_fw.hpp
SunnySrivastava1984: I will need some more insight on runtime and compile time argument here.

jinuthomas (changes requested):
still not convinced by the return , as nowhere is it mentioned that positive numbers are a failure , either check equals 0for success or negative for failure , will help if they make a change tomorrow for adding positive values for some reason.

I can live with the trow , but my question is is it that important , I am ok with having it there also

src/pldm_fw.cpp
PriyangaRamasamy: resolved

src/pldm_fw.cpp
PriyangaRamasamy: done

PriyangaRamasamy:
Changes made in [476b4f9](https://github.com/ibm-openbmc/ibm-panel/commit/476b4f9f3c421b1bbfd7cd0e3dd2f92e69c71b3a)
1. Updated the error check portion to check on negative errors (cross checked with libpldm apis)
2. Removed meson option to initiate transport layer. Currently we have only one way to initiate transport layer which is using mctp demux deamon running on BMC. Thus removed the unused AF MCTP logic to open and close transport layer.
3. Instance ID data base is not initiated at constructor.. It will get initiated on demand.

include/pldm_fw.hpp
```diff
@@ -43,6 +46,17 @@ class PldmFramework
     /** Host mctp eid */
     static constexpr auto mctpEid = (types::Byte)9;
 
+    static constexpr types::TerminusID tid = mctpEid;
+
+    /** @brief PLDM instance ID database object used to get instance IDs
+     */
+    pldm_instance_db* pldmInstanceIdDb = nullptr;
```
souvik1914581: Why are we using raw pointers here instead of C++ smart pointers?

souvik1914581: Ok, I see this is dictated by `pldm_instance_db_init_default` API.

PriyangaRamasamy: right

src/pldm_fw.cpp
```diff
@@ -7,33 +7,140 @@
...
+            std::to_string(rc));
+    }
+
+    // Get Instance ID
+    pldm_instance_id_t instanceID = 0;
+
+    rc = pldm_instance_id_alloc(pldmInstanceIdDb, tid, &instanceID);
+    if (rc == -EAGAIN)
+    {
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
souvik1914581, line 36: How did you arrive on this sleep time of 100ms before retry?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,140 @@
...
+        throw std::runtime_error(
+            "Call to pldm_instance_db_init_default failed with return code " +
+            std::to_string(rc));
+    }
+
+    // Get Instance ID
+    pldm_instance_id_t instanceID = 0;
+
+    rc = pldm_instance_id_alloc(pldmInstanceIdDb, tid, &instanceID);
+    if (rc == -EAGAIN)
```
souvik1914581, line 34: Why are we comparing `rc` with `-EAGAIN` (negative `EAGAIN`) ? `EAGAIN` would be some positive integer value defined in `errno.h` right?

souvik1914581: Ok, the `pldm_instance_id_alloc` API negates the error numbers before returning.

include/pldm_fw.hpp
```diff
@@ -81,12 +95,32 @@ class PldmFramework
 
     /**
      * @brief Get instance ID
-     * This api returns the instance id by making a dbus call to GetInstanceId
-     * api of Pldm. Instance id is to uniquely identify a message packet. This
-     * id is generated by pldm.
+     * This api returns the instance id by making a dbus call to libpldm
+     * api. Instance id is to uniquely identify a message packet.
+     * @return pldm_instance_id_t instance id.
```
souvik1914581, line 104: Nit: `@throw` missing

src/pldm_fw.cpp
```diff
@@ -7,33 +7,140 @@
...
+{
+    int rc = pldm_instance_id_free(pldmInstanceIdDb, tid, instanceID);
+    if (rc < 0)
+    {
+        std::cerr << "pldm_instance_id_free failed to free id = "
+                  << static_cast<int>(instanceID)
+                  << " of tid = " << static_cast<int>(tid) << " rc = " << rc
+                  << std::endl;
+    }
+
```
souvik1914581, line 67: Should we try to destroy instance ID DB if freeing instance ID has failed?

src/pldm_fw.cpp
PriyangaRamasamy: yes

src/pldm_fw.cpp
PriyangaRamasamy: yes. instance id DB is created whenever new instance ID is allocated and destroyed whenever the instance ID is destroyed

src/pldm_fw.cpp
PriyangaRamasamy: it's a random pick to retry the instance id allocation after 0.1 second

src/pldm_fw.cpp
```diff
@@ -7,33 +7,140 @@
...
+    // Destroy the instance ID DB
+    rc = pldm_instance_db_destroy(pldmInstanceIdDb);
+    if (rc < 0)
+    {
+        std::cout << "pldm_instance_db_destroy failed, rc =" << rc << "\n";
+
+        if (pldmInstanceIdDb)
+        {
+            pldmInstanceIdDb = nullptr;
```
souvik1914581: We should call `free(pldmInstanceIdDb)` to destroy the instance ID DB object "manually", and then do `pldmInstanceIdDb = nullptr`.

include/pldm_fw.hpp
```diff
@@ -81,12 +95,32 @@ class PldmFramework
...
+    int openPldm();
+
+    /** @brief Opens the MCTP socket for sending and receiving messages.
+     *
+     * @return file descriptor to perform MCTP demux based communication.
+     */
+    int openMctpDemuxTransport();
+
+    /** @brief Close the PLDM file */
+    void pldmClose();
```
souvik1914581: The open method is `openPldm()`, so to be consistent, this method should be named`closePldm()`

src/pldm_fw.cpp
```diff
@@ -7,33 +7,140 @@
...
+        std::cerr
+            << "Call to pldm_transport_mctp_demux_map_tid failed with rc = "
+            << rc << std::endl;
+        pldmClose();
+        throw FunctionFailure(
+            "Call to pldm_transport_mctp_demux_map_tid failed with rc = " +
+            std::to_string(rc));
+    }
+
+    pldmTransport = pldm_transport_mctp_demux_core(mctpDemux);
```
souvik1914581: Missing return value check.

src/pldm_fw.cpp
```diff
@@ -7,33 +7,140 @@
...
+
+    struct pollfd pollfd;
```
souvik1914581, line 122: Nit: Suggest using a variable name which is different than the datatype, incase someone does a `typedef struct pollfd pollfd;` in the future, it may lead to compilation errors.

Also, as a good practice, suggest initializing `pollfd` members with some invalid values.

src/pldm_fw.cpp
```diff
@@ -7,33 +7,140 @@
...
     }
-    return instanceId;
+    return pollfd.fd;
+}
+
+void PldmFramework::pldmClose()
+{
+    pldm_transport_mctp_demux_destroy(mctpDemux);
+    mctpDemux = nullptr;
+    pldmTransport = nullptr;
```
souvik1914581: Does `libpldm` take ownership of memory pointed to by `pldmTransport` and free it?

include/pldm_fw.hpp
PriyangaRamasamy: this doesn't impact the code quality. both gives same meaning. am ok with pldmClose

include/pldm_fw.hpp
```diff
@@ -81,12 +95,32 @@ class PldmFramework
...
+     * @brief Free PLDM requester instance id
+     * @param[in] pdrs - Panel PDR data.
+     */
+    void freeInstanceID(pldm_instance_id_t instanceID);
+
+    /**
+     * @brief setup PLDM transport for sending and receiving messages
      *
-     * @return one byte instance id.
+     * @return file descriptor
```
souvik1914581: Some more description will be helpful.
`file descriptor` to do what?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,140 @@
...
+    {
+        std::cerr << "open: pldmTransport already setup!" << std::endl;
+        throw FunctionFailure("Required host dump action via PLDM is not "
+                              "allowed due to openPldm failed");
+    }
+
+    fd = openMctpDemuxTransport();
+    if (fd < 0)
+    {
+        auto e = errno;
```
souvik1914581: What is the use of reading `errno` here?

I see `openMctpDemuxTransport()` throws exception incase of any failure, so the `if(fd < 0)` block will not execute.

In case`openMctpDemuxTransport()` does not encounter any internal failure, and returns `-1`,  I do not see it setting `errno` anywhere.

src/pldm_fw.cpp
```diff
@@ -113,77 +220,78 @@ types::PldmPacket
...
+            freeInstanceID(instance);
+            throw std::runtime_error(
+                "pldm:SetStateEffecterStates request message empty");
+        }
 
-    int fd = pldm_open();
-    if (fd == -1)
+        // Open PLDM socket based communication
+        int fd = openPldm();
+        if (fd == -1)
```
souvik1914581: I see `openPldm()` throws exception incase of any errors, so this `if (fd == -1)` block is dead code.

src/pldm_fw.cpp
PriyangaRamasamy: no. we should handle it

src/pldm_fw.cpp
```diff
@@ -113,77 +220,78 @@ types::PldmPacket
...
+        for (const auto i : packet)
+        {
+            std::cout << std::setfill('0') << std::setw(2) << std::hex << (int)i
+                      << " ";
+        }
+        std::cout << std::endl;
+
+        freeInstanceID(instance);
+        pldmClose();
+        if (rc < 0)
```
souvik1914581: Why is the return value being checked down here and not immediately after `pldm_transport_send_msg()` call?

src/pldm_fw.cpp
PriyangaRamasamy: pollfd initialisation is taken care by **pldm_transport_mctp_demux_init_pollfd** API. so no need of default initialisation from panel application side as we cannot decide any default value.

this is how pollfd struct looks like.
struct pollfd {
	int fd;	       /* file descriptor */
	short events;  /* requested events */
	short revents; /* returned events */
};

PriyangaRamasamy: >Nit: Suggest using a variable name which is different than the datatype, incase someone does a typedef struct pollfd pollfd; in the future, it may lead to compilation errors.

sure will consider this point. right now we don't have any conflicts with type and variable name. will keep it unchanged

src/pldm_fw.cpp
PriyangaRamasamy: pldmTransport is the struct. if I use is_empty() to check it will always tell its not empty as it is_empty will check if struct has atleast one data member.. it doesn;t validate the value of that data member..
 
struct pldm_transport {
	const char *name;
	uint8_t version;
	pldm_requester_rc_t (*recv)(struct pldm_transport *transport,
				    pldm_tid_t *tid, void **pldm_resp_msg,
				    size_t *msg_len);
	pldm_requester_rc_t (*send)(struct pldm_transport *transport,
				    pldm_tid_t tid, const void *pldm_msg,
				    size_t msg_len);
	int (*init_pollfd)(struct pldm_transport *transport,
			   struct pollfd *pollfd);
};

PriyangaRamasamy: so no need of return value check here

src/pldm_fw.cpp
PriyangaRamasamy: why we need both free and pldmInstanceIdDb = nullptr. ? 

using free() is not required as if pldmInstanceIdDB is already destroyed (or not created itself) then free() will throw undefined behaviour.
pldmInstanceIdDb = nullptr; will do the deallocation of memory. so we are good here

src/pldm_fw.cpp
PriyangaRamasamy: >What is the use of reading errno here?

errno is set by C library when the system call fails. so we need errno if we receive fd as -1.



>I do not see it setting errno anywhere.

what do you mean?


>I see openMctpDemuxTransport() throws exception incase of any failure, so the if(fd < 0) block will not execute.

see there are two things to consider here.. one is openMctpDemuxTransport() API has a flow to open a mctp socket.. if there is any issue in the flow then we throw exception..

if flow succeded there is a chance that we may receive filedescriptor as -1. and that is why we have these checks.. what do you mean by if(fd < 0) block will not execute?

src/pldm_fw.cpp
PriyangaRamasamy: because before throwing the exception saying SEND FAILED, we are doing certain things, viz., 
1. we are capturing the data  packet sent on console 
2. we are freeing up the instance ID
3. we are closing the PLDM socket.

these are required before we throw and exit

src/pldm_fw.cpp
PriyangaRamasamy: I'll remove this check

include/pldm_fw.hpp
PriyangaRamasamy: it is the file descriptor of the socket established by PLDM for communication between BMC and host. will update in the file

include/pldm_fw.hpp
PriyangaRamasamy: done

src/pldm_fw.cpp
souvik1914581: Ok, I see `pldm_instance_db_destroy()` API does a `free()` on the `pldmInstanceIdDb` pointer.
Ideally, the API itself should have done `pldmInstanceIdDb = nullptr`.

src/pldm_fw.cpp
```diff
@@ -7,33 +7,131 @@
...
-        std::cerr << e.what() << std::endl;
-        throw FunctionFailure("pldm: call to GetInstanceId failed.");
+        std::cout << "pldm_instance_db_destroy failed, rc =" << rc << "\n";
+
+        if (pldmInstanceIdDb)
+        {
+            pldmInstanceIdDb = nullptr;
+            std::cout << "\n Destroyed pldm instance ID DB object manually"
+                      << std::endl;
+        }
```
souvik1914581, line 74: `if(rc >= 0)` case implies memory pointed to by `pldmInstanceIdDb` has been freed by `pldm_instance_db_destroy()`, the `pldmInstanceIdDb` pointer needs to be set to `nullptr` here.

src/pldm_fw.cpp
souvik1914581: I meant we should check for `nullptr` return value.

src/pldm_fw.cpp
souvik1914581: Ok.

src/pldm_fw.cpp
souvik1914581: I do not see a `free(pldmTransport)` anywhere in our code?

src/pldm_fw.cpp
PriyangaRamasamy: so irrespective of rc, we need to set pldmInstanceIdDb to nullptr.

souvik1914581: Correct, because `pldmInstanceIdDb` becomes a dangling pointer after `pldm_instance_db_destroy()` and should be set to `nullptr` to prevent any future dereference.

src/pldm_fw.cpp
PriyangaRamasamy: it is there in libpldm code 
https://github.com/openbmc/libpldm/blob/779e9dbdcfaf8e131b6c35edbc4e057aab52a65a/src/transport/mctp-demux.c#L273

souvik1914581: Ok. Got it.

souvik1914581 (changes requested):
`nullptr` return value check missing in one place and dead code removal pending.
Otherwise looks fine to me.

src/pldm_fw.cpp
PriyangaRamasamy: done

souvik1914581 (approved):

include/pldm_fw.hpp
```diff
@@ -43,6 +46,17 @@ class PldmFramework
     /** Host mctp eid */
     static constexpr auto mctpEid = (types::Byte)9;
 
+    static constexpr types::TerminusID tid = mctpEid;
```
SunnySrivastava1984: A one liner to explain what it is or what it will be used for will be handy?

include/pldm_fw.hpp
```diff
@@ -43,6 +46,17 @@ class PldmFramework
...
+
+    /** @brief PLDM instance ID database object used to get instance IDs
```
SunnySrivastava1984: We don't use @brief in case of member variable. Should be removed for consistency?

include/pldm_fw.hpp
```diff
@@ -43,6 +46,17 @@ class PldmFramework
...
+     */
+    pldm_instance_db* pldmInstanceIdDb = nullptr;
+
+    /** pldm transport instance  */
+    pldm_transport* pldmTransport = nullptr;
+
+    pldm_transport_mctp_demux* mctpDemux = nullptr;
```
SunnySrivastava1984: Same here, a one liner.

include/pldm_fw.hpp
```diff
@@ -81,12 +95,40 @@ class PldmFramework
...
+     *
+     * @return file descriptor of the socket established by PLDM for
+     * communication between BMC and host.
+     *
+     * @throw Exception when BMC to HOST(via PLDM) connection establishment
+     * fails.
+     */
+    int openPldm();
+
+    /** @brief Opens the MCTP socket for sending and receiving messages.
```
SunnySrivastava1984: @brief should be in the next line.

src/pldm_fw.cpp
```diff
@@ -7,33 +7,135 @@
 #include <libpldm/platform.h>
 #include <libpldm/pldm.h>
 #include <libpldm/state_set.h>
+#include <poll.h>
```
SunnySrivastava1984, line 10: Just curious, if the meson check 
if cxx.has_header('poll.h') fails, still this will compile?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,135 @@
...
+            std::to_string(rc));
+    }
+
+    // Get Instance ID
+    pldm_instance_id_t instanceID = 0;
+
+    rc = pldm_instance_id_alloc(pldmInstanceIdDb, tid, &instanceID);
+    if (rc == -EAGAIN)
+    {
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
SunnySrivastava1984, line 36: Not sure, how much change can happen under 100 milli seconds. Should the wait be for some considerable time? Else what is the point re-trying?

src/pldm_fw.cpp
```diff
@@ -113,77 +215,71 @@ types::PldmPacket
...
-    int fd = pldm_open();
-    if (fd == -1)
+        if (packet.empty())
+        {
+            throw std::runtime_error(
+                "pldm:SetStateEffecterStates request message empty");
+        }
+
+        // Open PLDM socket based communication
+        [[maybe_unused]] int fd = openPldm();
```
SunnySrivastava1984: Am I missing something? I don't see "fd" being used anywhere. Why are we declaring it?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,135 @@
...
     {
-        std::cerr << e.what() << std::endl;
-        throw FunctionFailure("pldm: call to GetInstanceId failed.");
+        std::cerr << "open: pldmTransport already setup!" << std::endl;
+        throw FunctionFailure("Required host dump action via PLDM is not "
+                              "allowed due to openPldm failed");
     }
-    return instanceId;
+
+    fd = openMctpDemuxTransport();
```
SunnySrivastava1984: Could have returned the function call itself. Don't see a need of another local variable.

SunnySrivastava1984 (changes requested):

include/pldm_fw.hpp
PriyangaRamasamy: its the host terminus ID which is already there in include/types.hpp..
For more details refer PLDM and libpldm documentation.

include/pldm_fw.hpp
PriyangaRamasamy: removed

include/pldm_fw.hpp
PriyangaRamasamy: added

include/pldm_fw.hpp
PriyangaRamasamy: done

src/pldm_fw.cpp
PriyangaRamasamy: check fails ? did you mean if the check returns false?

if it returns false, compilation will not fail.

src/pldm_fw.cpp
PriyangaRamasamy: ideally instance id allocation should not fail.. But for some reason if libpldm fails to get instance id for it's users - it returns EAGAIN asking us to try querying again. I don't find any rule in libpldm asking us to retry after a delay. This delay is for our safer side. 100ms is working fine till now. I would suggest if we hit any issues then we can modify this delay.

src/pldm_fw.cpp
PriyangaRamasamy: modified

src/pldm_fw.cpp
PriyangaRamasamy: we need it for error logging.. will update

include/pldm_fw.hpp
SunnySrivastava1984: Then why is it there for "pldmInstanceIdDb", "pldmTransport" and "mctpDemux".
Can't that be referred from somewhere?

It was asked fro consistency through out the file.

src/pldm_fw.cpp
SunnySrivastava1984: My concern was if you are checking "if cxx.has_header('poll.h')" then you imply that there is a possibility of compiler not having poll.h.
So if it is actually missing then in that case how will "#include <poll.h>" compile? From where it will get included?

src/pldm_fw.cpp
SunnySrivastava1984: Ok.

src/pldm_fw.cpp
```diff
@@ -7,33 +7,133 @@
...
+    }
+
+    std::cout << "Got instanceId: " << static_cast<int>(instanceID)
+              << " from PLDM eid: " << static_cast<int>(tid) << std::endl;
+    return instanceID;
+}
+
+void PldmFramework::freeInstanceID(pldm_instance_id_t instanceID)
+{
+    int rc = pldm_instance_id_free(pldmInstanceIdDb, tid, instanceID);
```
SunnySrivastava1984: What happens if "pldmInstanceIdDb" is a null pointer?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,133 @@
...
+    }
+
+    // Destroy the instance ID DB
+    rc = pldm_instance_db_destroy(pldmInstanceIdDb);
+    if (rc < 0)
+    {
+        std::cout << "pldm_instance_db_destroy failed, rc =" << rc << "\n";
+    }
+
+    if (pldmInstanceIdDb)
```
SunnySrivastava1984: Do we need both ways of destruction? Why can't we just close it manually and we are done?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,133 @@
...
+{
+    int rc = pldm_transport_mctp_demux_init(&mctpDemux);
+    if (rc < 0)
+    {
+        std::cerr << "Call to pldm_transport_mctp_demux_init failed with rc = "
+                  << rc << std::endl;
+        throw FunctionFailure("Failed to init MCTP demux transport");
+    }
+
+    rc = pldm_transport_mctp_demux_map_tid(mctpDemux, tid, tid);
```
SunnySrivastava1984: one of the tid is supposed to be "pldm tid" and another is supposed to be "mctp eid" if I am not wrong. 
Now since the value and type is same this will work.
But the variable that is being used here does not state that at all. Expect a better nomenclature here.

src/pldm_fw.cpp
```diff
@@ -7,33 +7,133 @@
...
+    if (rc < 0)
```
SunnySrivastava1984: Is this check valid?
"pldm_transport_mctp_demux_map_tid" will return any negative value?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,133 @@
...
+
+    if (!pldmTransport)
+    {
+        throw FunctionFailure(
+            "openMctpDemuxTransport: Failed to get pldm_transport object.");
+    }
+
+    struct pollfd pollfd;
+    rc = pldm_transport_mctp_demux_init_pollfd(pldmTransport, &pollfd);
+    if (rc < 0)
```
SunnySrivastava1984: Can "pldm_transport_mctp_demux_init_pollfd" return a negative value? I don't see this API returning negative value.
https://github.com/openbmc/libpldm/blob/779e9dbdcfaf8e131b6c35edbc4e057aab52a65a/src/transport/mctp-demux.c#L74

Am I referring to a wrong place? If so please correct me, else
I don't see a need for return type being checked for negative value.

meson.build
```diff
@@ -34,6 +34,10 @@ systemd_system_unit_dir = systemd.get_variable('systemdsystemunitdir')
 
 service_file = 'service_files/com.ibm.panel.service'
 
+if cxx.has_header('poll.h')
+  add_project_arguments('-DPLDM_HAS_POLL=1', language: 'cpp')
```
SunnySrivastava1984: Why are we adding this? Not able to understand the use case.

SunnySrivastava1984 (changes requested):

include/pldm_fw.hpp
PriyangaRamasamy: done

src/pldm_fw.cpp
PriyangaRamasamy: yes we need to tell libpldm that this instance of DB got destroyed. and also we need to set null in our class member variable.

src/pldm_fw.cpp
PriyangaRamasamy: The issue is not with nomenclature. Here we intentionally send tid in the place of eid.
Why because EID is the entity ID which is a specific term used in PLDM. DMTF has no way to distinguish between TID and EID. 

PLDM will build the infrastructure to get the correct TID from the EID. Until then we use eid as tid in EID-TID mapping.

src/pldm_fw.cpp
PriyangaRamasamy: it returns 0 now. but its return type is int and on error cases it may return negative numbers in future when there is any change in library functions. That's why I have this check here.

src/pldm_fw.cpp
PriyangaRamasamy: it returns 0 now. but its return type is int and on error cases it may return negative numbers in future when there is any change in library functions. That's why I have this check here.

src/pldm_fw.cpp
PriyangaRamasamy: pldm_instance_id_free() returns -EINVAL

src/pldm_fw.cpp
PriyangaRamasamy: right, I added an ifdef check before including poll.h

meson.build
PriyangaRamasamy: libpldm has this config check to use polling APIs. since we are using the poll apis directly from libpldm, we need this config check

include/pldm_fw.hpp
```diff
@@ -81,12 +97,41 @@ class PldmFramework
...
+     * @return pldm_instance_id_t instance id.
      *
-     * @return one byte instance id.
+     * @throw Exception when failed to allocate instance ID.
+     */
+    pldm_instance_id_t getInstanceID();
+
+    /**
+     * @brief Free PLDM requester instance id
+     * @param[in] pdrs - Panel PDR data.
```
souvik1914581: `@param[in]` does not match with actual parameter

src/pldm_fw.cpp
```diff
@@ -8,32 +8,151 @@
...
-    try
+    if (rc < 0)
+    {
+        throw std::runtime_error(
+            "Call to pldm_instance_db_init_default failed with return code " +
+            std::to_string(rc));
+    }
+
+    // PLDM_INST_ID_MAX is 32 (0 to 31). Initialising with an invalid/unused ID.
+    pldm_instance_id_t instanceID = 100;
```
souvik1914581: We should initialize with `PLDM_INST_ID_MAX + 1` to make sure `instanceID` here is always invalid.

src/pldm_fw.cpp
```diff
@@ -8,32 +8,151 @@
...
+              << " from PLDM eid: " << static_cast<int>(tid) << std::endl;
+    return instanceID;
+}
+
+void PldmFramework::freeInstanceID(pldm_instance_id_t instanceID)
+{
+    if (pldmInstanceIdDb)
+    {
+        // PLDM_INST_ID_MAX is 32 (0 to 31)
+        if (instanceID <= 31)
```
souvik1914581: Should be `instanceID <= PLDM_INST_ID_MAX - 1`
I see `PLDM_INST_ID_MAX` in `instance-id.c` file of libpldm. We should have a macro defined in our code which is a copy of the library's macro.

src/pldm_fw.cpp
```diff
@@ -8,32 +8,151 @@
...
+        {
+            int rc = pldm_instance_id_free(pldmInstanceIdDb, tid, instanceID);
```
souvik1914581: `pldm_instance_id_free` can return `-EAGAIN`. We should add re-try logic here as well.

src/pldm_fw.cpp
```diff
@@ -113,77 +232,75 @@ types::PldmPacket
 void PldmFramework::sendPanelFunctionToPhyp(
     const types::FunctionNumber& funcNumber)
 {
-    types::PdrList pdrs =
-        utils::getPDR(phypTerminusID, frontPanelBoardEntityId,
-                      stateIdToEnablePanelFunc, "FindStateEffecterPDR");
+    // PLDM_INST_ID_MAX is 32 (0 to 31). Initialising with an invalid/unused ID.
+    pldm_instance_id_t instance = 100;
```
souvik1914581: Should be initialized with `PLDM_INST_ID_MAX`.

src/pldm_fw.cpp
```diff
@@ -8,32 +8,151 @@
...
+}
+
+void PldmFramework::pldmClose()
+{
+    if (mctpDemux)
+    {
+        pldm_transport_mctp_demux_destroy(mctpDemux);
+        mctpDemux = nullptr;
+    }
+
```
souvik1914581, line 142: A comment here which says `pldm_transport_mctp_demux_destroy()` takes care of de-allocating memory pointed to by `pldmTransport` would be useful.

src/pldm_fw.cpp
```diff
@@ -8,32 +8,151 @@
...
+    std::cout << "Got instanceId: " << static_cast<int>(instanceID)
+              << " from PLDM eid: " << static_cast<int>(tid) << std::endl;
+    return instanceID;
+}
+
+void PldmFramework::freeInstanceID(pldm_instance_id_t instanceID)
+{
+    if (pldmInstanceIdDb)
```
souvik1914581: nit: `if(!pldmInstanceIdDb) { return; } ` would be more readable.
We wouldn't need to ensure all the subsequent code is inside the `if(pldmInstanceIdDb)` block.

souvik1914581 (changes requested):

src/pldm_fw.cpp
PriyangaRamasamy: done

src/pldm_fw.cpp
PriyangaRamasamy: done

src/pldm_fw.cpp
PriyangaRamasamy: done

src/pldm_fw.cpp
PriyangaRamasamy: done

src/pldm_fw.cpp
PriyangaRamasamy: done

src/pldm_fw.cpp
PriyangaRamasamy: done

include/pldm_fw.hpp
PriyangaRamasamy: done

meson.build
```diff
@@ -55,72 +64,57 @@ panel_app_a = static_library(
     'src/bus_monitor.cpp',
     'src/executor.cpp',
     'src/pldm_fw.cpp',
-    include_directories: 'include'
+    include_directories: 'include',
```
souvik1914581, line 67: The `,` got added by `clang` ?

souvik1914581: The `,` got added by `clang` ?

meson.build
```diff
@@ -55,72 +64,57 @@ panel_app_a = static_library(
...
 )
 panel_tool_a = static_library(
-     'ibm_dbus_call_a',
-     'tools/src/dbus_call.cpp',
-     include_directories: ['include', 'tools/include']
+    'ibm_dbus_call_a',
+    'tools/src/dbus_call.cpp',
+    include_directories: ['include', 'tools/include'],
```
souvik1914581: The `,` got added by `clang` ?

souvik1914581 (commented):
Few minor comments, other than that looks good to me.

meson.build
PriyangaRamasamy: yes

meson.build
PriyangaRamasamy: yes

souvik1914581 (approved):

meson.build
```diff
@@ -20,29 +20,38 @@ phosphor_dbus_interfaces = dependency('phosphor-dbus-interfaces')
...
-cxx.get_supported_arguments([
-'-DBOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT',
-'-DBOOST_ASIO_DISABLE_THREADS',
-'-DBOOST_NO_RTTI',
-'-DBOOST_NO_TYPEID',
-'-DBOOST_ALLOW_DEPRECATED_HEADERS'
-]),
-language : 'cpp')
-add_global_arguments('-Wno-psabi', language : ['c', 'cpp'])
+    cxx.get_supported_arguments(
```
SunnySrivastava1984: Why are we taking clang formatting of meson file via this PR?
Shouldn't this be a separate PR to fix clang? This change unnecessarily makes both the changes dependent.

src/pldm_fw.cpp
```diff
@@ -8,32 +8,161 @@
...
+            std::cerr << "pldm_instance_id_free failed to free id = "
+                      << static_cast<int>(instanceID)
+                      << " of tid = " << static_cast<int>(tid) << " rc = " << rc
+                      << std::endl;
+        }
+    }
+
+    // Destroy the instance ID DB
+    int rc = pldm_instance_db_destroy(pldmInstanceIdDb);
+    if (rc < 0)
```
SunnySrivastava1984: Why is this being checked for less than 0? Called API will always return 0.

src/pldm_fw.cpp
SunnySrivastava1984: How is it known that in future it will always have negative value on an error case?
Unless that is guaranteed, this still is not a valid code.

src/pldm_fw.cpp
SunnySrivastava1984: How is it known that in future it will always have negative value on an error case?
Unless that is guaranteed, this still is not a valid code.

src/pldm_fw.cpp
```diff
@@ -113,77 +242,75 @@ types::PldmPacket
...
+            prepareSetEffecterReq(pdrs, instance, funcNumber);
+
+        if (packet.empty())
+        {
+            throw std::runtime_error(
+                "pldm:SetStateEffecterStates request message empty");
+        }
+
+        // Open PLDM socket based communication
+        int fd = openPldm();
```
SunnySrivastava1984: Not sure if we should continue if the fd is "0" as the "openMctpDemuxTransport" failed? or is it ok to continue?

SunnySrivastava1984 (changes requested):

src/pldm_fw.cpp
PriyangaRamasamy: On failure libpldm apis will return value less than 0.

The below link has the requester's expected return value on success and failure. https://github.com/openbmc/libpldm/blob/5192e2e26f26ffbbf4256429ea5ae42237e011cc/include/libpldm/pldm.h#L15

meson.build
PriyangaRamasamy: anyways this changes are dependent on clang fix.. that's why I pushed in the same PR.

src/pldm_fw.cpp
PriyangaRamasamy: https://github.com/openbmc/libpldm/blob/5192e2e26f26ffbbf4256429ea5ae42237e011cc/include/libpldm/pldm.h#L15
the above shows that the requester error codes are negative integers.

src/pldm_fw.cpp
PriyangaRamasamy: https://github.com/openbmc/libpldm/blob/5192e2e26f26ffbbf4256429ea5ae42237e011cc/include/libpldm/pldm.h#L15
the above shows that the requester error codes are negative integers.

src/pldm_fw.cpp
PriyangaRamasamy: libpldm returns fd as 0 when PLDM_HAS_POLL config is not declared. So keeping panel in sync with libpldm.

https://github.com/openbmc/libpldm/blob/5192e2e26f26ffbbf4256429ea5ae42237e011cc/src/transport/transport.c#L27C19-L27C23

meson.build
PriyangaRamasamy: but with a different commit.

PriyangaRamasamy: moved to separate commit

include/const.hpp
```diff
@@ -123,5 +123,8 @@ static const uint8_t LEN_OF_RAW_HEX_WORD = 4;
 static const uint8_t WORDS_PER_DISPLAY = 4;
 
 static const std::string defaultHexWordValue = "00000000";
+
+// Maximum valid instance IDs' which can be created by libpldm
+static const auto PLDM_INST_ID_MAX = 32;
```
souvik1914581, line 128: nit: this can be `constexpr`

souvik1914581: nit: this can be `constexpr`

src/pldm_fw.cpp
```diff
@@ -7,33 +7,154 @@
...
+    }
+
+    if (rc != 0)
+    {
+        throw FunctionFailure(
+            "Call to pldm_instance_id_alloc failed with return code " +
+            std::to_string(rc));
+    }
+
+    std::cout << "Got instanceId: " << static_cast<int>(instanceID)
```
souvik1914581: Do these traces add any value or are they just there for debugging purpose?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,154 @@
...
+        if (rc != 0)
+        {
+            std::cerr << "pldm_instance_id_free failed to free id = "
+                      << static_cast<int>(instanceID)
+                      << " of tid = " << static_cast<int>(tid) << " rc = " << rc
+                      << std::endl;
+        }
+    }
+
+    // Destroy the instance ID DB
```
souvik1914581, line 77: Couple of points here:
1. If `instanceID >= PLDM_INST_ID_MAX`, should we proceed to destroy the instance ID Database?
2.  The doxygen of `freeInstanceID` tells `Free PLDM requester instance id`, but it is also destroying the instance ID Database, which is another functionality. I think "destroying the instance ID Database" should be moved to another method.

src/pldm_fw.cpp
```diff
@@ -7,33 +7,154 @@
...
+    {
+        std::cout << "pldm_instance_db_destroy failed, rc =" << rc << "\n";
+    }
+
+    pldmInstanceIdDb = nullptr;
+    std::cout << "\n Destroyed pldm instance ID DB object manually"
+              << std::endl;
+}
+
+int PldmFramework::openPldm()
```
souvik1914581: The only logic inside `openPldm()` currently is to check if pldmTransport is already setup and then it calls `openMctpDemuxTransport()`.

Can we move "check if pldmTransport is already setup" inside `openMctpDemuxTransport()` and eliminate `openPldm()` ?

src/pldm_fw.cpp
```diff
@@ -113,77 +234,75 @@ types::PldmPacket
...
+        std::cout << "Panel function " << static_cast<int>(funcNumber)
+                  << ". Data packet sent to pldm: ";
+        for (const auto i : packet)
+        {
+            std::cout << std::setfill('0') << std::setw(2) << std::hex << (int)i
+                      << " ";
+        }
+        std::cout << std::endl;
+
+        if (rc != 0)
```
souvik1914581: This `rc` should be compared with `PLDM_REQUESTER_SUCCESS` incase the `pldm_requester_rc_t` enum values in the library change in future. 

Also, the doxygen of `pldm_transport_send_msg` mentions `errno` maybe set. The `errno` should also be captured as part of the `exception.what()`

souvik1914581 (changes requested):

src/pldm_fw.cpp
PriyangaRamasamy: it's there for debugging purpose

src/pldm_fw.cpp
PriyangaRamasamy: 1.  I can add a check if (pldmInstanceIdDb != nullptr) before destroying.
2. the idea here is to have single function to allocate and deallocate instance id and instance id DB. so that the caller won't miss freeing up instance ID DB. I can update the doxygen and api name..

src/pldm_fw.cpp
PriyangaRamasamy: in future we will have to get rid of openMctpDemuxTransport and move to AF Mctp Transport layer 
ref: https://github.com/ibm-openbmc/ibm-panel/pull/213/files#diff-42320549c8b243f96275cd488c49a08636a92e9c8c7b178351fd5d75911f1091R66

so let's have a wrapper openPldm as a common API - which can chose between either mctp demux or af mctp demux or any other layer for opening the communication socket.

src/pldm_fw.cpp
PriyangaRamasamy: done

src/pldm_fw.cpp
souvik1914581: Ok, I am fine with this.

src/pldm_fw.cpp
souvik1914581: Alright. Ok with this.

src/pldm_fw.cpp
souvik1914581: As discussed, I am fine with this.

souvik1914581 (approved):
LGTM

include/pldm_fw.hpp
```diff
@@ -43,6 +46,19 @@ class PldmFramework
     /** Host mctp eid */
     static constexpr auto mctpEid = (types::Byte)9;
 
+    /** Terminus ID to which the packet has to be sent */
+    static constexpr types::TerminusID tid = mctpEid;
```
jinuthomas: why is tid needed ?

include/pldm_fw.hpp
```diff
@@ -43,6 +46,19 @@ class PldmFramework
     /** Host mctp eid */
     static constexpr auto mctpEid = (types::Byte)9;
```
jinuthomas: can you rename this to hostEid

include/pldm_fw.hpp
```diff
@@ -43,6 +46,19 @@ class PldmFramework
...
 
+    /** Terminus ID to which the packet has to be sent */
+    static constexpr types::TerminusID tid = mctpEid;
+
+    /** PLDM instance ID database object used to get instance IDs */
+    pldm_instance_db* pldmInstanceIdDb = nullptr;
```
jinuthomas: nit: pldmInstanceDb

include/pldm_fw.hpp
```diff
@@ -43,6 +46,19 @@ class PldmFramework
     /** Host mctp eid */
     static constexpr auto mctpEid = (types::Byte)9;
 
+    /** Terminus ID to which the packet has to be sent */
+    static constexpr types::TerminusID tid = mctpEid;
+
+    /** PLDM instance ID database object used to get instance IDs */
```
jinuthomas: instance database object not instance ID database object

include/pldm_fw.hpp
```diff
@@ -43,6 +46,19 @@ class PldmFramework
...
+    pldm_instance_db* pldmInstanceIdDb = nullptr;
+
+    /** pldm transport instance  */
+    pldm_transport* pldmTransport = nullptr;
+
+    /** PLDM MCTP demux object which is required to establish MCTP socket based
+     * communication. */
+    pldm_transport_mctp_demux* mctpDemux = nullptr;
```
jinuthomas: should this not be named pldmSocket ?

jinuthomas: would be better to have it generic so we do not change later

include/pldm_fw.hpp
```diff
@@ -81,12 +97,46 @@ class PldmFramework
...
+     * @throw Exception when BMC to HOST(via PLDM) connection establishment
+     * fails.
+     */
+    int openPldm();
+
+    /**
+     * @brief Opens the MCTP socket for sending and receiving messages.
      *
-     * @return one byte instance id.
+     * @return file descriptor to perform MCTP demux based communication.
```
jinuthomas: just file descriptor of socket would be good

src/pldm_fw.cpp
jinuthomas: I have a question on this based on your last sentence , since you said that this working fine , wanted to know how did you test this condition to be working fine ?

include/pldm_fw.hpp
jinuthomas: Also which is the host here hosboot or PHYP ?

jinuthomas (changes requested):
I do understand that you have used the variables similar to what it was earlier , Also I do understand that you coded looking at the PLDM header, I would like to suggest some extra changes , if you do find that not good, maybe I can do the changes on top of your changes. I am sorry but I felt a bit confused and thought it would be better to have a discussion with you and the team to see if we can generalize this more and save us from knowing what PLDM does internally from the code.

include/pldm_fw.hpp
PriyangaRamasamy: done

include/pldm_fw.hpp
PriyangaRamasamy: host here is phyp.

include/pldm_fw.hpp
PriyangaRamasamy: terminus id is the pldm terminology of pointing the destination at which the packet has to be sent.
eid is the mctp id of the host.. 
https://github.com/openbmc/pldm/blob/4bf3ed840f13c31720ffa7c255942f3836402f4f/common/transport.cpp#L18

SunnySrivastava1984 (approved):

src/pldm_fw.cpp
```diff
@@ -7,33 +7,143 @@
...
 #include <sdbusplus/asio/object_server.hpp>
 
 namespace panel
 {
-types::Byte PldmFramework::getInstanceID()
+pldm_instance_id_t PldmFramework::getInstanceID()
 {
-    types::Byte instanceId = 0;
-    auto bus = sdbusplus::bus::new_default();
+    // Fetch Instance ID DB
```
jinuthomas, line 20: nit: // Fetch instance DB

src/pldm_fw.cpp
```diff
@@ -7,33 +7,143 @@
...
+            std::to_string(rc));
+    }
+
+    // Initialising with an invalid/unused ID.
+    pldm_instance_id_t instanceID = constants::PLDM_INST_ID_MAX + 1;
+
+    rc = pldm_instance_id_alloc(pldmInstanceDb, tid, &instanceID);
+    if (rc == -EAGAIN)
+    {
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
jinuthomas, line 36: not sure if this is even valid .
How did you test that 100 milliseconds is a good value ?
If cannot be tested , do recommend to remove this code. 
What does PLDM library tell if this fails ?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,143 @@
...
+    }
+
+    if (rc != 0)
+    {
+        throw FunctionFailure(
+            "Call to pldm_instance_id_alloc failed with return code " +
+            std::to_string(rc));
+    }
+
+    return instanceID;
```
jinuthomas, line 47: Do we need to check if the value is valid i.e within the MAX alloted value

src/pldm_fw.cpp
```diff
@@ -7,33 +7,143 @@
...
-    try
+    if (rc != 0)
+    {
+        throw std::runtime_error(
+            "Call to pldm_instance_db_init_default failed with return code " +
+            std::to_string(rc));
+    }
+
+    // Initialising with an invalid/unused ID.
+    pldm_instance_id_t instanceID = constants::PLDM_INST_ID_MAX + 1;
```
jinuthomas, line 31: why a +1 here as per your later comment in code I see the below
// constants::PLDM_INST_ID_MAX is 32 (0 to 31)

include/pldm_fw.hpp
```diff
@@ -41,7 +44,20 @@ class PldmFramework
     // TODO: <https://github.com/ibm-openbmc/ibm-panel/issues/57>
     // use PLDM defined header file to refer following constants.
     /** Host mctp eid */
-    static constexpr auto mctpEid = (types::Byte)9;
+    static constexpr auto hostEid = (types::Byte)9;
```
jinuthomas, line 47: why do you need this variable ?

src/pldm_fw.cpp
```diff
@@ -113,77 +223,76 @@ types::PldmPacket
...
+        // Open PLDM socket based communication
+        int fd = openTransport();
+
+        pldm_tid_t pldmTID = static_cast<pldm_tid_t>(hostEid);
+        auto rc = pldm_transport_send_msg(pldmTransport, pldmTID, packet.data(),
+                                          packet.size());
 
-    int fd = pldm_open();
-    if (fd == -1)
+        std::cout << "Panel function " << static_cast<int>(funcNumber)
```
jinuthomas, line 262: move this to the failure path

include/types.hpp
```diff
@@ -85,6 +85,9 @@ using PendingAttributesItemType =
     std::pair<std::string, std::tuple<std::string, AttributeValueType>>;
 using PendingAttributesType = std::vector<PendingAttributesItemType>;
 
+/** @brief Host terminus ID. */
+using TerminusID = uint8_t;
```
jinuthomas, line 89: why is this needed ?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,143 @@
...
+    {
+        return;
+    }
+
+    // constants::PLDM_INST_ID_MAX is 32 (0 to 31)
+    if (instanceID < constants::PLDM_INST_ID_MAX)
+    {
+        int rc = pldm_instance_id_free(pldmInstanceDb, tid, instanceID);
+
+        if (rc == -EAGAIN)
```
jinuthomas, line 62: same question as in alloc 
is this even needed and how do we justify the 100 ms

src/pldm_fw.cpp
```diff
@@ -7,33 +7,143 @@
...
+        {
+            std::this_thread::sleep_for(std::chrono::milliseconds(100));
+            rc = pldm_instance_id_free(pldmInstanceDb, tid, instanceID);
+        }
+
+        if (rc != 0)
+        {
+            std::cerr << "pldm_instance_id_free failed to free id = "
```
jinuthomas, line 70: Why a print here and no error log ?

src/pldm_fw.cpp
```diff
@@ -113,77 +223,76 @@ types::PldmPacket
 void PldmFramework::sendPanelFunctionToPhyp(
     const types::FunctionNumber& funcNumber)
 {
-    types::PdrList pdrs =
-        utils::getPDR(phypTerminusID, frontPanelBoardEntityId,
-                      stateIdToEnablePanelFunc, "FindStateEffecterPDR");
+    // Initialising with invalid/unused ID
+    pldm_instance_id_t instance = constants::PLDM_INST_ID_MAX + 1;
```
jinuthomas, line 227: why the +1 ?

src/pldm_fw.cpp
```diff
@@ -7,33 +7,143 @@
...
+        if (rc != 0)
+        {
+            std::cerr << "pldm_instance_id_free failed to free id = "
+                      << static_cast<int>(instanceID)
+                      << " of tid = " << static_cast<int>(tid) << " rc = " << rc
+                      << std::endl;
+        }
+    }
+
+    // Destroy the instance ID DB
```
jinuthomas, line 77: nit: instance DB

jinuthomas (commented):
have a bit of concerns in the error path

PR #224, opened 2025-02-20
Enable SRC functions on progress code events

PriyangaRamasamy:
As functions 11, 12, 13 are to display SRCs & hexwords of any type (say progress SRC/ PEL SRC), whichever SRC event comes first these functions has to be enabled.

In the existing design the functions 11, 12, 13 are enabled only when PEL SRC events are received.

This commit enables the SRC functions 11, 12 and 13 even when progress code events are received by ibm-panel application.

Test:
Tested on P10 rainier to see even if PEL is not found, on occurrence of progress code events, function 11, 12, 13 are enabled.

### DIFF
--- a/src/bus_monitor.cpp
+++ b/src/bus_monitor.cpp
@@ -257,8 +257,8 @@ void PELListener::setPelRelatedFunctionState(
 
     if (!functionStateEnabled)
     {
-        // these functions needs to be enabled only once when first
-        // PEL of desired severity is received.
+        // these functions needs to be enabled only once when first PEL of
+        // desired severity is received.
         functionStateEnabled = true;
         list.emplace_back(11);
         list.emplace_back(12);
@@ -496,6 +496,18 @@ void BootProgressCode::progressCodeCallBack(sdbusplus::message_t& msg)
                     }
                 }
                 executor->storeSRCAndHexwords(hexWordsWithSRC);
+
+                // Enable functions when progress code is received
+                types::FunctionalityList list;
+                list.reserve(3);
+
+                // these functions needs to be enabled once when progress code
+                // is received
+                list.emplace_back(11);
+                list.emplace_back(12);
+                list.emplace_back(13);
+
+                stateManager->enableFunctonality(list);
             }
         }
         else
--- a/include/bus_monitor.hpp
+++ b/include/bus_monitor.hpp
@@ -163,10 +163,13 @@ class BootProgressCode
      * @param[in] con - Bus connection.
      * @param[in] execute - pointer to Executor.
      */
-    BootProgressCode(std::shared_ptr<Transport> transport,
-                     std::shared_ptr<sdbusplus::asio::connection> con,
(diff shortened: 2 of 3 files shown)


src/bus_monitor.cpp
```diff
@@ -257,8 +257,8 @@ void PELListener::setPelRelatedFunctionState(
 
     if (!functionStateEnabled)
     {
-        // these functions needs to be enabled only once when first
-        // PEL of desired severity is received.
+        // these functions needs to be enabled only once either when progress
```
jinuthomas: don't see a change required here in the message , as this function is for PEL only

src/bus_monitor.cpp
```diff
@@ -496,6 +496,18 @@ void BootProgressCode::progressCodeCallBack(sdbusplus::message_t& msg)
                     }
                 }
                 executor->storeSRCAndHexwords(hexWordsWithSRC);
+
+                // Enable functions when progress code is received
```
jinuthomas, line 500: here also 11 needs to be enabled for progress code based on what you said,
Also 12 and 13 needs to be enabled only if there is hex words.

src/bus_monitor.cpp
```diff
@@ -496,6 +496,18 @@ void BootProgressCode::progressCodeCallBack(sdbusplus::message_t& msg)
...
+                types::FunctionalityList list;
+                list.reserve(3);
+
+                // these functions needs to be enabled once either when progress
```
jinuthomas: this comment needs to be rewritten as this function is for progress code and not PEL

jinuthomas (changes requested):

src/bus_monitor.cpp
PriyangaRamasamy: right.. this function is for PEL only, but the comment there is specific to function 11,12,13.
the comment says functions 11, 12 13 has to be enabled only once when first PEL is recevied - which I slightly modified.

src/bus_monitor.cpp
PriyangaRamasamy: okay got it.. will make the change

src/bus_monitor.cpp
PriyangaRamasamy: ok.. you want it like that.. fine will modify both the comments

src/bus_monitor.cpp
jinuthomas: The comment should be based on the function , need not be based on the function number

src/bus_monitor.cpp
jinuthomas: Please always comment based on the function only.

PriyangaRamasamy: sure

src/bus_monitor.cpp
PriyangaRamasamy: ok

jinuthomas (approved):

PR #225, opened 2025-02-20
Bug fix:Function 12&13 scrolls off the display

PriyangaRamasamy:
This commit fixes the code bug which triggers function 12 and 13 to scroll off the display.

Test:
Before this change, display looks like:
ibm-panel[1079]: L1 : 03031003100003100000
ibm-panel[1079]: L2 :

After this change, display looks like:
ibm-panel[1079]: L1 : 03100000
ibm-panel[1079]: L2 :

### DIFF
--- a/src/bus_monitor.cpp
+++ b/src/bus_monitor.cpp
@@ -493,6 +493,8 @@ void BootProgressCode::progressCodeCallBack(sdbusplus::message_t& msg)
                             << static_cast<int>(hexWordArray.at(arrayLoop++));
 
                         hexWordsWithSRC += convert.str();
+                        // clear the buffer
+                        convert.str("");
                     }
                 }
                 executor->storeSRCAndHexwords(hexWordsWithSRC);


jinuthomas (changes requested):
did not understand this as I did not see the 4 hex words
normally in the BMC system you will need to see two words per line as standard. so I miss what you are trying to do here , nor did not understand the commit message

src/bus_monitor.cpp
```diff
@@ -493,6 +493,8 @@ void BootProgressCode::progressCodeCallBack(sdbusplus::message_t& msg)
                             << static_cast<int>(hexWordArray.at(arrayLoop++));
 
                         hexWordsWithSRC += convert.str();
+                        // clear the buffer
+                        convert.str("");
```
PriyangaRamasamy, line 497: Loop at line 480 : 
starts from 8 (0 indexing) ( because we skip the first 8 bytes(index 0-7) as they are the header)
byte 8 - 11 will be the first hex word 
byte 12 - 15 second hex word
byte 16-19 third
byte 20- 23 fourth hex word


Each hex word is of 4bytes.. we are converting it into 8 bytes by setting the width of each byte to 2. ( at line 491)

Eg: HEX WORD = 3D06
for loop at line 489:
 iteration 1 :  value of the buffer convert = 03
                      value of hexWordsWithSRC = 03
 iteration 2:  value of the buffer convert = 030D
                      value of hexWordsWithSRC = 03 + 030D
  iteration 3:  value of the buffer convert = 030D00
                       value of hexWordsWithSRC = 03030D + 030D00
   iteration 4:  value of the buffer convert = 030D0006
                        value of hexWordsWithSRC = 03030D030D00 + 030D0006
                        
  Expected value of hexWordsWithSRC = 030D0006
  Due to bug the value of hexWordsWithSRC = 03030D030D00030D0006
   
   
at line 497 -> we convert it into 03 
03d00060

PriyangaRamasamy:
> did not understand this as I did not see the 4 hex words normally in the BMC system you will need to see two words per line as standard. so I miss what you are trying to do here , nor did not understand the commit message

I've quote the reply at code section for better understanding

jinuthomas (approved):

PR #226, opened 2025-02-20
Function 11,12,13:Conflicting space delimiter

PriyangaRamasamy:
Function 11 is to display primary SRC and function 12 and 13 will display the hex words 2,3,4,5 and 6,7,8,9 respectively. Existing code uses space as the delimiter to distinguish between primary SRC and hex words. Where in some cases primary SRC itself has a space in it for which function 11 execution fails to display it due to which functions 12 and 13 displays incorrect output.

Changes in function 11:
This commit replaces the check that displays primary SRC based on the space With the size of the SRC (which is 8 bytes in length).

Changes in function 12 and 13:
Start string split from hexwords and not from SRC.

Test:

ibm-panel[840]: L1 : 11
ibm-panel[840]: L2 :

ibm-panel[840]: L1 : Linux pp
ibm-panel[840]: L2 :

ibm-panel[840]: L1 : 12
ibm-panel[840]: L2 :

ibm-panel[840]: L1 : 03100000
ibm-panel[840]: L2 :

ibm-panel[840]: L1 : 13
ibm-panel[840]: L2 :

ibm-panel[840]: L1 :
ibm-panel[840]: L2 :

### DIFF
--- a/include/const.hpp
+++ b/include/const.hpp
@@ -106,5 +106,22 @@ static const types::PICFRUPathMap bootFailPIC = {
     {blueridge2s4uIM, systemDbusObj},
     {blueridge1s4uIM, systemDbusObj},
     {fujiIM, everBMCObj}};
+
+static const uint8_t FUNCTION_12 = 12;
+static const uint8_t FUNCTION_13 = 13;
+
+/** PEL SRC data structure has the following sub sections:
+ * 1. 8 byte header.
+ * 2. HEX words each of length 4 bytes.
+ * 3. 32 bytes ascii character string.
+ *
+ * Refer PEL SRC data structure for more details.
+ */
+static const uint8_t FIRST_HEX_WORD_START_OFFSET = 8;
+static const uint8_t LAST_HEX_WORD_START_OFFSET = 36;
+static const uint8_t LEN_OF_RAW_HEX_WORD = 4;
+static const uint8_t WORDS_PER_DISPLAY = 4;
+
+static const std::string defaultHexWordValue = "00000000";
 } // namespace constants
 } // namespace panel
--- a/include/executor.hpp
+++ b/include/executor.hpp
@@ -313,6 +313,15 @@ class Executor
      */
     void sendFuncNumToPhyp(const types::FunctionNumber& funcNumber);
 
+    /**
+     * @brief API to display SRC functions 12 and 13
+     * This is a common API which can be used to display extended hex SRC words
+     * in function 12 and function 13 on request.
+     *
+     * @param[in] function - Function number (12/13).
+     */
+    void displayHexWords(const uint8_t function);
+
     /*Transport class object*/
     std::shared_ptr<Transport> transport;
 
@@ -337,8 +346,27 @@ class Executor
     /* OS IPL mode state */
     bool osIplMode = false;
 
-    /* SRC and HEX words */
+    /**
+     * Primary SRC and 8 extended hex SRC words are saved in the string
+     * latestSrcAndHexwords. Primary SRC and each extended hex SRC words is of
+     * length 8 bytes. A space is provided as a delimiter between every 8 bytes.
+     */
     std::string latestSrcAndHexwords;
+
+    /** Offsets to find the SRC words stored in the string
+     * latestSrcAndHexwords.*/
+    const uint8_t primarySrcOffset = 0;
+    const uint8_t offsetOfHexWord2 = 9;
+    const uint8_t offsetOfHexWord3 = 18;
+    const uint8_t offsetOfHexWord4 = 27;
+    const uint8_t offsetOfHexWord5 = 36;
+    const uint8_t offsetOfHexWord6 = 45;
+    const uint8_t offsetOfHexWord7 = 54;
+    const uint8_t offsetOfHexWord8 = 63;
+    const uint8_t offsetOfHexWord9 = 72;
+
+    /** Word length */
+    const uint8_t wordLength = 8;
 
     /* To keep track if the execute request is from external or not */
     bool isExternallyTriggered = false;
--- a/src/bus_monitor.cpp
+++ b/src/bus_monitor.cpp
@@ -461,60 +461,53 @@ void BootProgressCode::progressCodeCallBack(sdbusplus::message_t& msg)
                 return;
             }
 
-            // To detect if there is a need to save SRCs and hexwords in func 11
-            // to 13, check for array data filled with space.
-            if (hexWordArray.at(0) != 0x20)
+            // store the primary SRC.
+            std::string hexWordsWithSRC =
+                std::string(byteArray.begin(), byteArray.end());
+
+            /* Skip first 8 byte header to get the hex word start offset */
+            for (auto hexWordStartOffset =
+                     constants::FIRST_HEX_WORD_START_OFFSET;
+                 hexWordStartOffset <= constants::LAST_HEX_WORD_START_OFFSET;)
             {
-                // store the SRC.
-                std::string hexWordsWithSRC =
-                    std::string(byteArray.begin(), byteArray.end());
-
-                // 4th byte will return number of valid hex words.
-                types::Byte validHexWords = hexWordArray.at(3);
-
-                // Ignoring the first 8 bytes from the array as those are some
-                // header related data not HEX words. Hex words are represented
-                // as 4 bytes so read 4*validHexWords bytes to read all the
-                // hexwords.
                 std::ostringstream convert;
-                for (size_t arrayLoop = 8; arrayLoop <= (4 * validHexWords);)
+
+                // Add a space in between each extended hex SRC word
+                hexWordsWithSRC += " ";
+
+                // Read 4 bytes and convert each byte to hex byte of width 2
+                for (size_t hexWordLoop = 0;
+                     hexWordLoop < constants::LEN_OF_RAW_HEX_WORD;
+                     ++hexWordLoop)
                 {
-                    // clear any previous data.
-                    convert.str("");
-
-                    hexWordsWithSRC += " ";
-
-                    // From this offet read 4 bytes, convert them to HEX and
-                    // then append to final string of hexwords and SRC.
-                    for (size_t hexWordLoop = 0; hexWordLoop < 4; ++hexWordLoop)
-                    {
-                        convert
-                            << std::setfill('0') << std::setw(2) << std::hex
-                            << static_cast<int>(hexWordArray.at(arrayLoop++));
-
-                        hexWordsWithSRC += convert.str();
-                        // clear the buffer
-                        convert.str("");
-                    }
+                    convert << std::setfill('0') << std::setw(2) << std::hex
+                            << static_cast<int>(
+                                   hexWordArray.at(hexWordStartOffset++));
                 }
-                executor->storeSRCAndHexwords(hexWordsWithSRC);
-
-                // Enable functions when progress code is received
-                types::FunctionalityList list;
-                list.reserve(3);
-
-                // these functions needs to be enabled once when progress code
-                // is received
-                list.emplace_back(11);
-                list.emplace_back(12);
-                list.emplace_back(13);
-
-                stateManager->enableFunctonality(list);
+                // Append the 8 byte extended hex SRC word to the string
+                hexWordsWithSRC += convert.str();
             }
+
+            executor->storeSRCAndHexwords(hexWordsWithSRC);
+
+            // Enable functions when progress code is received
+            types::FunctionalityList list;
+            list.reserve(3);
+
+            // these functions needs to be enabled once when progress code
+            // is received
+            list.emplace_back(11);
+            list.emplace_back(12);
+            list.emplace_back(13);
+
+            stateManager->enableFunctonality(list);
         }
         else
         {
-            std::cerr << "Progress code Data error" << std::endl;
+            std::cerr
+                << "Unable to get progress code from "
+                   "xyz.openbmc_project.State.Boot.Raw.Value D-bus property."
+                << std::endl;
         }
     }
 }
--- a/src/executor.cpp
+++ b/src/executor.cpp
@@ -290,27 +290,17 @@ void Executor::execute20()
 
 void Executor::execute11()
 {
-    if (!latestSrcAndHexwords.empty())
+    if (latestSrcAndHexwords.size() >= wordLength)
     {
-        // find the first space to get response code
-        auto pos = latestSrcAndHexwords.find_first_of(" ");
-
-        // length of src data need to be 8
-        if (pos != std::string::npos && pos == 8)
-        {
-            utils::sendCurrDisplayToPanel((latestSrcAndHexwords).substr(0, pos),
-                                          std::string{}, transport);
-        }
-        else
-        {
-            std::cerr << "Invalid srcData received = " << latestSrcAndHexwords
-                      << std::endl;
-        }
-        return;
+        utils::sendCurrDisplayToPanel(
+            (latestSrcAndHexwords).substr(primarySrcOffset, wordLength),
+            std::string{}, transport);
     }
-
-    // TODO: Decide what needs to be done in this case.
-    std::cerr << "Error getting SRC data" << std::endl;
+    else
+    {
+        std::cerr << "Invalid SRC data received = " << latestSrcAndHexwords
+                  << std::endl;
+    }
 }
 
 static std::string getEthLocPort(const std::string& macAddr)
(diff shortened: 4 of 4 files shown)


src/executor.cpp
```diff
@@ -290,27 +290,16 @@ void Executor::execute20()
...
+                                      std::string{}, transport);
+    }
+    else
+    {
+        std::cerr << "Invalid srcData received = " << latestSrcAndHexwords
+                  << std::endl;
     }
-
-    // TODO: Decide what needs to be done in this case.
-    std::cerr << "Error getting SRC data" << std::endl;
```
jinuthomas, line 313: why is this removed ?

src/executor.cpp
```diff
@@ -565,16 +554,21 @@ void Executor::execute12()
...
+    if (!latestSrcAndHexwords.empty() && latestSrcAndHexwords.size() > 9)
     {
-        std::vector<std::string> src;
-        boost::split(src, latestSrcAndHexwords, boost::is_any_of(" "));
+        // Extract hex words from the string which has both primary src and
+        // hexwords Remove first 9 elements which has primary SRC (8 bytes) and
+        // a space (1 byte).
+        auto latestHexWordsString = latestSrcAndHexwords.substr(9);
+
+        std::vector<std::string> hexWordsArray;
```
jinuthomas: why not get the first set of hex words 2-5 here ?
also why do you not check if all the 4 hex words are there or not

src/executor.cpp
```diff
@@ -565,16 +554,21 @@ void Executor::execute12()
...
+        boost::split(hexWordsArray, latestHexWordsString,
+                     boost::is_any_of(" "));
 
-        const auto size = std::min(src.size(), (size_t)5);
-        // ignoring the first hexword as that will be SRC.
-        for (size_t i = 1; i < size; ++i)
+        const auto size = std::min(hexWordsArray.size(), (size_t)4);
```
jinuthomas: maybe a little optimization here will be good

src/executor.cpp
```diff
@@ -590,16 +584,22 @@ void Executor::execute13()
...
+    if (!latestSrcAndHexwords.empty() && latestSrcAndHexwords.size() > 9)
     {
-        std::vector<std::string> src;
-        boost::split(src, latestSrcAndHexwords, boost::is_any_of(" "));
+        // Extract hex words from the string which has both primary src and
+        // hexwords Remove first 9 elements which has primary SRC (8 bytes) and
+        // a space (1 byte).
+        auto latestHexWordsString = latestSrcAndHexwords.substr(9);
+
+        std::vector<std::string> hexWordsArray;
```
jinuthomas: same a above you could have cut the second set of hex words here if any by checking the remaining size and then could have used that directly

jinuthomas (changes requested):
I do agree with you that the data saved may not be needing a " "(space) as delimiter. you can maybe change that also so that the numbering can be easier to set than adding for space

src/executor.cpp
PriyangaRamasamy: >why not get the first set of hex words 2-5 here ?

hex words 2 to 5 are separated by a space ( while forming the string in bus_monitor.cpp).
line 565 - is splitting the hexwords based on space.

PriyangaRamasamy: >why do you not check if all the 4 hex words are there or not

it is not mandatory to have all 4 hex words. Some SRCs will have 1 or 2 hex words only..

src/executor.cpp
PriyangaRamasamy: one of the test cases is:
hexWordsArray can have all the hexwords 2,3,4,5,6,7,8,9.. in that we need only first 4 hexwords to get printed on function12.. that's why we have std::min(total array size , 4)..

what is your thought in optimising the line 568?

src/executor.cpp
PriyangaRamasamy: yeah, as I told above we have space separated hex words.. so we cannot cut the hexwords 6,7,8,9 based on the size of hexwords :)

src/executor.cpp
PriyangaRamasamy: The two if else got replaced with single if else.. so std::cerr at line 300 is enough to display the invalid src.

src/executor.cpp
```diff
@@ -565,16 +554,21 @@ void Executor::execute12()
...
+        boost::split(hexWordsArray, latestHexWordsString,
+                     boost::is_any_of(" "));
```
jinuthomas: change this

jinuthomas (changes requested):

include/executor.hpp
```diff
@@ -313,6 +313,15 @@ class Executor
...
     void sendFuncNumToPhyp(const types::FunctionNumber& funcNumber);
 
+    /**
+     * @brief API to display SRC functions 12 and 13
+     * This is a common API which can be used to display extended hex SRC words
+     * in function 12 and function 13 on request.
+     *
+     * @param[in] function - Function number (12/13).
+     */
+    void displaySRCFunction(const uint8_t function);
```
branupama: camel case: displaySrcFunction ?

include/executor.hpp
```diff
@@ -337,12 +346,42 @@ class Executor
...
+    ** Word 1 is the primary SRC.
+    ** Words 2 to 9 are the 8 extended hex SRC words.
+    ** Each word is of 8 byte length. Total size of latestSrcAndHexwords string
+    will be [8 * 9 SRC words + 8 spaces] = 80.
+    */
     std::string latestSrcAndHexwords;
 
+    /** Offsets to find the SRC words stored in the string
+     * latestSrcAndHexwords.*/
+    const uint8_t primarySRCOffset = 0;
```
branupama: camel case: primarySrcOffset

src/bus_monitor.cpp
```diff
@@ -461,60 +461,59 @@ void BootProgressCode::progressCodeCallBack(sdbusplus::message_t& msg)
...
+             *extended hex SRC words which we are interested in starts from
+             *word 2. To calculate the count of valid extended hex SRC words,
+             *remove the count for first word (which is a part of the header).
+             **/
+            types::Byte validHexWords = hexWordArray.at(3) - 1;
+
+            /* Skip first 8 byte header to get the hex word start offset */
+            uint8_t hexWordStartOffset = 8;
+
+            while (validHexWords-- > 0)
```
branupama: as we know already out of 72 bytes, start 8 bytes is HEADER and last 32 bytes are ASCII string.
Can we keep 8 directly here as a valid hex words count and can avoid while loop ?

src/bus_monitor.cpp
```diff
@@ -461,60 +461,59 @@ void BootProgressCode::progressCodeCallBack(sdbusplus::message_t& msg)
...
-                        convert
-                            << std::setfill('0') << std::setw(2) << std::hex
-                            << static_cast<int>(hexWordArray.at(arrayLoop++));
-
-                        hexWordsWithSRC += convert.str();
-                        // clear the buffer
-                        convert.str("");
-                    }
+                // Read 4 bytes and convert each byte to hex byte of width 2
+                for (size_t hexWordLoop = 0; hexWordLoop < 4; ++hexWordLoop)
```
branupama: can we keep 4 in constants, as it represents each hexword's size in number of bytes.

src/executor.cpp
```diff
@@ -290,27 +290,18 @@ void Executor::execute20()
 
 void Executor::execute11()
 {
-    if (!latestSrcAndHexwords.empty())
+    if (!latestSrcAndHexwords.empty() &&
```
branupama: no need to check emptiness, as size checked with  wordLength is sufficient.

include/executor.hpp
```diff
@@ -337,12 +346,42 @@ class Executor
...
+    const uint8_t offsetOfHexWord5 = 36;
+    const uint8_t offsetOfHexWord6 = 45;
+    const uint8_t offsetOfHexWord7 = 54;
+    const uint8_t offsetOfHexWord8 = 63;
+    const uint8_t offsetOfHexWord9 = 72;
+
+    const uint8_t maxLengthOfSavedWords = 80;
+
+    /** SRC Word length */
+    const uint8_t wordLength = 8;
```
branupama, line 369: can we specify it as srcWordLength = 8 ?
and in comment we can update SRC word size in bytes.

src/executor.cpp
```diff
@@ -290,27 +290,18 @@ void Executor::execute20()
...
-                      << std::endl;
-        }
-        return;
+        utils::sendCurrDisplayToPanel(
+            (latestSrcAndHexwords).substr(primarySRCOffset, wordLength),
+            std::string{}, transport);
+    }
+    else
+    {
+        std::cerr << "Invalid srcData received = " << latestSrcAndHexwords
```
branupama: nit: SRC word

src/executor.cpp
```diff
@@ -290,27 +290,18 @@ void Executor::execute20()
 
 void Executor::execute11()
```
branupama, line 291: should we add function11 in displaySRCFunction(), as it is already handing func 12, 13. 
As funs 11, 12, 13 are related. ?

include/executor.hpp
```diff
@@ -337,12 +346,42 @@ class Executor
...
+    string latestSrcAndHexwords.
+
+    ------------------------------------------------------------------------
+    WORD1 | space | WORD2 | space | WORD3 | space | WORD4 | space | WORD5 |
+    space | WORD6 | space | WORD7 | space | WORD8 | space | WORD9 |
+    ------------------------------------------------------------------------
+
+    ** Word 1 is the primary SRC.
+    ** Words 2 to 9 are the 8 extended hex SRC words.
+    ** Each word is of 8 byte length. Total size of latestSrcAndHexwords string
```
branupama: 2-9 word are 4 byte length each.

include/executor.hpp
```diff
@@ -337,12 +346,42 @@ class Executor
...
+    /**
+    PRIMARY SRC and extended hex SRC words are saved in this pattern in the
+    string latestSrcAndHexwords.
+
+    ------------------------------------------------------------------------
+    WORD1 | space | WORD2 | space | WORD3 | space | WORD4 | space | WORD5 |
+    space | WORD6 | space | WORD7 | space | WORD8 | space | WORD9 |
+    ------------------------------------------------------------------------
+
+    ** Word 1 is the primary SRC.
```
branupama: and is of 8 byte length.

include/executor.hpp
```diff
@@ -337,12 +346,42 @@ class Executor
...
+    ** Words 2 to 9 are the 8 extended hex SRC words.
+    ** Each word is of 8 byte length. Total size of latestSrcAndHexwords string
+    will be [8 * 9 SRC words + 8 spaces] = 80.
```
branupama: 8 SRC words. and total should it be 48 bytes ?
Is it necessary to give this total size in comment ?

branupama (changes requested):

src/bus_monitor.cpp
```diff
@@ -461,60 +461,59 @@ void BootProgressCode::progressCodeCallBack(sdbusplus::message_t& msg)
...
+            std::string hexWordsWithSRC =
+                std::string(byteArray.begin(), byteArray.end());
+
+            /** Byte 4 in the Base SRC structure gives the total number of hex
+             *words. As per SPEC, word 1 is used as part of the header. The
+             *extended hex SRC words which we are interested in starts from
+             *word 2. To calculate the count of valid extended hex SRC words,
+             *remove the count for first word (which is a part of the header).
+             **/
+            types::Byte validHexWords = hexWordArray.at(3) - 1;
```
jinuthomas: why is valid hex words checked. 
I had mentioned not to use it

include/executor.hpp
PriyangaRamasamy: 2 to 9 are 4 byte length when you receive from PostCode interface.
while saving in string latestSrcAndHexWords - we will have 8 bytes ( as those 4 bytes are converted into 2 digit hex values )

include/executor.hpp
PriyangaRamasamy: as i told above each word will be 8 bytes of length while saving to the string latestSrcHexwords. 8 * 9 + 8 = 80.

The computation is based on the size. so yes it is necessary to mention size in the comment section

include/executor.hpp
PriyangaRamasamy: this length is for src and hex words. each word is of 8 byte length. wordLength suits for both the cases

src/executor.cpp
PriyangaRamasamy: no we cannot have.

src/executor.cpp
PriyangaRamasamy: ack

src/bus_monitor.cpp
PriyangaRamasamy: ack

src/executor.cpp
```diff
@@ -561,51 +552,52 @@ void Executor::execute01()
...
+            output = std::vector<std::string>(
+                {latestSrcAndHexwords.substr(offsetOfHexWord6, wordLength),
+                 latestSrcAndHexwords.substr(offsetOfHexWord7, wordLength),
+                 latestSrcAndHexwords.substr(offsetOfHexWord8, wordLength),
+                 latestSrcAndHexwords.substr(offsetOfHexWord9, wordLength)});
         }
     }
 
-    // send blank display if string is empty
+    // send zeroes if string is empty
```
branupama: no need of this comment, as we already initialising it to zero in the beginning and there is comment already at line no 560.

branupama (changes requested):

include/executor.hpp
PriyangaRamasamy: is there any hard rule to write acronyms in camelcase in c++?

src/executor.cpp
PriyangaRamasamy: ack

src/bus_monitor.cpp
PriyangaRamasamy: removed valid hex word count check

src/bus_monitor.cpp
PriyangaRamasamy: removed it

include/executor.hpp
branupama: I got it.

include/executor.hpp
branupama: okay

include/executor.hpp
branupama: should we update the comment for the same ?

include/const.hpp
```diff
@@ -106,5 +106,24 @@ static const types::PICFRUPathMap bootFailPIC = {
...
+4 byte WORD6 |  4 byte WORD7 |  4 byte WORD8 |  4 byte WORD9 | 32 BYTES ASCII
+char string|
+----------------------------------------------------------------------------
+*
+* SRC structure has the following sub sections:
+* 1. 8 byte header  [OFFSET 0 to 7]
+* 2. HEX words each of length 4 bytes [OFFSET 8 to 39]
+* 3. 32 bytes ascii character string [OFFSET 40 to 71]
+*/
+static const uint8_t HEX_DATA_SEC_START_OFFSET = 8;
```
branupama: what is SEC ?

PriyangaRamasamy: section

include/executor.hpp
PriyangaRamasamy: not required. SRC word length is sufficient

include/const.hpp
jinuthomas: you just need to provide the spec name here not list as above

include/executor.hpp
```diff
@@ -337,12 +346,42 @@ class Executor
...
+    const uint8_t wordLength = 8;
+
     /* To keep track if the execute request is from external or not */
     bool isExternallyTriggered = false;
 
     types::ReturnStatus executionStatus = std::make_tuple(false, "", "");
 }; // class Executor
-} // namespace panel
\ No newline at end of file
+} // namespace panel
```
jinuthomas, line 376: I expect this change due to clang formatting

include/const.hpp
```diff
@@ -106,5 +106,24 @@ static const types::PICFRUPathMap bootFailPIC = {
...
+static const uint8_t HEX_DATA_SEC_END_OFFSET = 39;
```
jinuthomas: why 39 here ?

include/executor.hpp
jinuthomas: please fix this

src/executor.cpp
```diff
@@ -561,51 +551,51 @@ void Executor::execute01()
...
-    // Need to show blank spaces in case no srcData as function is enabled.
-    constexpr auto blankHexWord = "        ";
-    std::vector<std::string> output(4, blankHexWord);
+    displaySRCFunction(constants::FUNCTION_12);
+}
+
+void Executor::displaySRCFunction(const uint8_t function)
+{
+    // Need to show zeroes in case no srcData as function is enabled.
+    constexpr auto hexZeroes = "00000000";
```
jinuthomas: this should be in constants , 
as defaultHexWordValue

src/executor.cpp
```diff
@@ -561,51 +551,51 @@ void Executor::execute01()
...
+    std::vector<std::string> output(4, hexZeroes);
```
jinuthomas: use TOTAL_RAW_HEX_WORD instead of 4

src/executor.cpp
```diff
@@ -561,51 +551,51 @@ void Executor::execute01()
...
 
     if (!latestSrcAndHexwords.empty())
     {
-        std::vector<std::string> src;
-        boost::split(src, latestSrcAndHexwords, boost::is_any_of(" "));
+        auto sizeOfLatestSrcAndHexwords = latestSrcAndHexwords.size();
+        if (sizeOfLatestSrcAndHexwords < maxLengthOfSavedWords)
+        {
+            // appened zeroes at the end
```
jinuthomas: This should never happen , bail out here

include/const.hpp
```diff
@@ -106,5 +106,24 @@ static const types::PICFRUPathMap bootFailPIC = {
     {blueridge2s4uIM, systemDbusObj},
     {blueridge1s4uIM, systemDbusObj},
     {fujiIM, everBMCObj}};
+
+static const uint8_t FUNCTION_12 = 12;
```
jinuthomas, line 110: why 12 ?

include/const.hpp
```diff
@@ -106,5 +106,24 @@ static const types::PICFRUPathMap bootFailPIC = {
...
+static const uint8_t FUNCTION_13 = 13;
```
jinuthomas, line 111: why 13 ?

include/executor.hpp
jinuthomas: yes we do follow that nomenclature in the Team, please fix

include/executor.hpp
```diff
@@ -313,6 +313,15 @@ class Executor
      */
     void sendFuncNumToPhyp(const types::FunctionNumber& funcNumber);
 
+    /**
+     * @brief API to display SRC functions 12 and 13
```
jinuthomas, line 317: This displays the hex words not the SRC function. either 2..5 or 6..9. :)

jinuthomas (changes requested):

src/executor.cpp
PriyangaRamasamy: done

src/executor.cpp
PriyangaRamasamy: TOTAL_RAW_HEX_WORDS is 8. here we display only 4.
WORDS_PER_DISPLAY = 4 , this suits better..is that fine?

src/executor.cpp
PriyangaRamasamy: moved

include/executor.hpp
PriyangaRamasamy: done

include/executor.hpp
PriyangaRamasamy: changed the function name

include/const.hpp
PriyangaRamasamy: i modified to take last hex word start offset and not end offset

jinuthomas (approved):
looks ok for the time being

PR #236, opened 2025-04-07
Fix system dump issue from OP-Panel

branupama:
When the user initiates system dump from OP-Panel on the system using function 42, is not being triggered.
Reason for this is “panel” service is not sending required parameters to ‘Dump manager’ application which make Dbus call to fail.

This commit adds the code to fix the above issue in the panel.

### DIFF
--- a/src/executor.cpp
+++ b/src/executor.cpp
@@ -928,16 +928,18 @@ void Executor::execute08()
     utils::sendCurrDisplayToPanel("SHUTDOWN SERVER", "INITIATED", transport);
 }
 
-static void createDump(const sdbusplus::message::object_path& object)
+static void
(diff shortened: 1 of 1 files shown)


src/executor.cpp
```diff
@@ -952,8 +954,11 @@ void Executor::execute43()
 
 void Executor::execute42()
 {
+    std::vector<std::pair<std::string, std::variant<std::string, uint64_t>>>
```
RekhaAparna01: can this be defined as const reference?

branupama: done.

branupama:
jenkins run tests please

jinuthomas (approved):

PR #240, opened 2025-05-15
Update input device path (#221)

eddiejames:
As a result of kernel update there was a change in the input device path.
The commit pulls in the change to pick correct device path.


eddiejames:
Needed for https://jazz07.rchland.ibm.com:13443/jazz/web/projects/CSSD#action=com.ibm.team.workitem.viewWorkItem&id=703153

PR #241, opened 2025-08-28
Lcd panel: Fix "device not found" error #729129

branupama:
The Panel app fails with the error "device not found" after an AC cycle
when it starts before the vpd-manager.

The vpd-manager is responsible for enabling the LCD microcontroller
I2C GPIO pin. When the Panel app starts after vpd-manager, it can
directly access the LCD microcontroller without issues.

However, if the Panel app comes up earlier, the GPIO pin is not yet
enabled, and access to the LCD microcontroller fails.

This change fixes the issue by checking the LCD microcontroller GPIO
presence pin and enabling the I2C GPIO pin before accessing the
microcontroller.

### DIFF
--- a/include/utils.hpp
+++ b/include/utils.hpp
@@ -40,6 +40,16 @@ static types::PanelDataMap lcdDataMap = {
      {constants::fujiLcdDevPath, constants::devAddr,
       constants::fujiLcdDbusObj}}};
 
+static const types::GpioInfoMap gpioInfo = {
+    {"gpioPresence", {"RUSSEL_OPPANEL_PRESENCE_N", 0}},
+    {"setGpio", {"RUSSEL_FW_I2C_ENABLE_N", 0}}};
+
+static const std::vector<std::string> systemsNeedsI2cEnableForLcd = {
+    constants::rain2s2uIM,      constants::rain2s4uIM,
+    constants::rain1s4uIM,      constants::balconesIM,
+    constants::blueridge2s2uIM, constants::blueridge2s4uIM,
+    constants::blueridge1s4uIM};
+
 /** @brief Read inventory manager properties from dbus.
  * @param[in] service - Dbus service name
  * @param[in] obj - Dbus object to query for the property.
@@ -260,6 +270,69 @@ void sortPels(types::GetManagedObjects& listOfPels);
  */
 void filterPel(const types::GetManagedObjects& listOfPels,
                types::PelPathAndSRCList& finalListOFPELs);
+
+/**
+ * @brief API to check if device is present.
+ *
+ * This API checks whether the device is present or not by reading device's gpio
+ * presence pin value.
+ *
+ * @param[in] gpioPinName - GPIO pin name.
+ * @param[in] expectedPinState - GPIO pin value to check.
+ * @param[in,out] errorReadingPin -  Flag to set in case of any error occured
+ * while reading the gpio pin.
+ *
+ * @return - Returns true if device is present, false otherwise.
+ */
+bool isDevicePresent(const std::string& gpioPinName,
+                     const uint8_t& expectedPinState,
+                     bool& errorReadingPin) noexcept;
+
+/**
+ * @brief API to set GPIO pin with given value.
+ *
+ * @param gpioPinName[in] - GPIO pin name.
+ * @param pinValue[in] - GPIO pin value.
+ *
+ * @return - Returns true if able to set gpio pin, false otherwise.
+ */
+bool setGpioPin(const std::string& gpioPinName,
+                const uint8_t& pinValue) noexcept;
+
+/**
+ * @brief API to enable gpio access for the given device.
+ *
+ * This API reads the device's gpio presence pin, if found enabled or in case of
+ * any error, will try to set the gpio output pin.
+ *
+ * In case of read or set gpio pin fails, will try to directly read the
+ * microcontroller firmware version, if succeds then it is considered as i2c
+ * access is enabled for the given device.
+ *
+ * @param gpioInfo[in] - Map containing gpio presence and output pin
+ * information.
+ * @param devPath[in] - Device i2c path.
+ * @param devAddress[in] - Device address.
+ *
+ * @return - true if able to access the device, false otherwise.
+ */
+bool enableDeviceI2cAccess(const panel::types::GpioInfoMap& gpioInfo,
+                           const std::string& devPath,
+                           const uint8_t& devAddress) noexcept;
+
+/**
+ * @brief API to check if device is accessible.
+ *
+ * This API checks whether the device is accessible or not by reading the
+ * microcontroller firmware version.
+ *
+ * @param[in] devPath - Device path.
+ * @param[in] devAddress - Device address.
+ *
+ * @return - true if device is accessible, false otherwise.
+ */
+bool isDeviceAccessible(const std::string& devPath,
+                        const uint8_t& devAddress) noexcept;
 
 } // namespace utils
 } // namespace panel
--- a/meson.build
+++ b/meson.build
@@ -17,6 +17,11 @@ project(
 systemd = dependency('systemd')
 sdbusplus = dependency('sdbusplus')
 phosphor_dbus_interfaces = dependency('phosphor-dbus-interfaces')
+libgpiodcxx = dependency(
+    'libgpiodcxx',
+    default_options: ['bindings=cxx'],
+    version: '<1.7.0',
+)
 
 cxx = meson.get_compiler('cpp')
 add_project_arguments(
@@ -88,7 +93,12 @@ executable(
 executable(
     'ibm-panel',
     'src/panel_app_main.cpp',
-    dependencies: [sdbusplus, dependency('libpldm'), phosphor_dbus_interfaces],
+    dependencies: [
+        sdbusplus,
+        dependency('libpldm'),
+        phosphor_dbus_interfaces,
+        libgpiodcxx,
+    ],
     include_directories: ['include', 'logger/include'],
     install: true,
     link_with: [panel_app_a, logger],
@@ -108,6 +118,7 @@ if get_option('tests').enabled()
             gtest,
             dependency('libpldm'),
             phosphor_dbus_interfaces,
+            libgpiodcxx,
         ],
         include_directories: ['include'],
         link_with: [panel_app_a],
--- a/src/panel_app_main.cpp
+++ b/src/panel_app_main.cpp
@@ -4,6 +4,9 @@
 #include "const.hpp"
 #include "utils.hpp"
 
+#include <unistd.h>
+
+#include <algorithm>
 #include <exception>
 #include <iostream>
 #include <sdbusplus/asio/connection.hpp>
@@ -170,8 +173,22 @@ int main(int, char**)
              * change from false to true; but the transport key is still
              * true(unchanged). To maintain data accuracy get the "Present"
              * property from dbus and set the transport key again.*/
-            lcdPanel->setTransportKey(
-                panel::utils::getLcdPanelPresentProperty(imValue));
+            if (std::find(panel::utils::systemsNeedsI2cEnableForLcd.begin(),
+                          panel::utils::systemsNeedsI2cEnableForLcd.end(),
+                          imValue) !=
+                panel::utils::systemsNeedsI2cEnableForLcd.end())
+            {
+                if (panel::utils::enableDeviceI2cAccess(panel::utils::gpioInfo,
+                                                        lcdDevPath, lcdDevAddr))
+                {
+                    lcdPanel->setTransportKey(true);
+                }
+            }
+            else
+            {
+                lcdPanel->setTransportKey(
+                    panel::utils::getLcdPanelPresentProperty(imValue));
+            }
         }
         else
         {
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -5,6 +5,13 @@
 #include "i2c_message_encoder.hpp"
 
 #include <libpldm/platform.h>
+#include <linux/i2c-dev.h>
+
+#include <chrono>
+#include <gpiod.hpp>
+#include <thread>
+
+using namespace std::chrono_literals;
 
 namespace panel
 {
@@ -440,6 +447,194 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
     }
 
     return finalListOfFPELs;
+}
+
+bool isDevicePresent(const std::string& gpioPinName,
+                     const uint8_t& expectedPinState,
+                     bool& errorReadingPin) noexcept
+{
+    if (gpioPinName.empty())
+    {
+        return false;
+    }
+
+    std::string errMsg{};
+    auto retries = constants::RETRY_COUNT;
+
+    while (retries--)
+    {
+        try
+        {
+            gpiod::line l_presenceLine = gpiod::find_line(gpioPinName);
+
+            if (!l_presenceLine)
+            {
+                throw std::runtime_error("Couldn't find the GPIO line.");
+            }
+
+            l_presenceLine.request({"Read the presence line",
+                                    gpiod::line_request::DIRECTION_INPUT, 0});
+
+            errorReadingPin = false;
+            return (l_presenceLine.get_value() == expectedPinState);
+        }
+        catch (const std::exception& ex)
+        {
+            errorReadingPin = true;
+            errMsg = ex.what();
+        }
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+    }
+
+    if (errorReadingPin)
+    {
+        std::cerr << "Error occured while checking gpio presence pin value "
+                     "for the pin ["
+                  << gpioPinName << "], expected pin state[" << expectedPinState
+                  << "], error: " << errMsg << std::endl;
+    }
+
+    return true;
+}
+
+bool setGpioPin(const std::string& gpioPinName,
+                const uint8_t& pinValue) noexcept
+{
+    if (gpioPinName.empty())
+    {
+        return false;
+    }
+
+    auto retries = constants::RETRY_COUNT;
+    std::string errMsg{};
+
+    while (retries--)
+    {
+        try
+        {
+            gpiod::line l_outputLine = gpiod::find_line(gpioPinName);
+            if (l_outputLine)
+            {
+                l_outputLine.request({"Panel App Action",
+                                      ::gpiod::line_request::DIRECTION_OUTPUT,
+                                      0},
+                                     pinValue);
+                return true;
+            }
+        }
+        catch (const std::exception& ex)
+        {
+            errMsg = ex.what();
+        }
+
+        std::this_thread::sleep_for(std::chrono::seconds(1));
+    }
+
+    if (!errMsg.empty())
+    {
+        std::cerr << "Error occured while enabling gpio line[" << gpioPinName
+                  << "], error: " << errMsg << std::endl;
+    }
+    return false;
+}
+
+bool isDeviceAccessible(const std::string& devPath,
+                        const uint8_t& devAddress) noexcept
+{
+    int panelFileDescriptor = -1;
+    bool isReadSuccessful = false;
+    try
+    {
+        if ((panelFileDescriptor = open(devPath.data(), O_RDWR | O_NONBLOCK)) ==
+            -1) // open failure
+        {
+            throw std::runtime_error("open system call failed, error code: " +
+                                     std::string(std::strerror(errno)));
+        }
+        if (ioctl(panelFileDescriptor, I2C_SLAVE, devAddress) ==
+            -1) // access failure
+        {
+            throw std::runtime_error("ioctl call failed, error code: " +
+                                     std::string(std::strerror(errno)));
+        }
+
+        panel::types::Binary versionBuffer;
+        constexpr const size_t versionSize = 6;
+        versionBuffer.resize(versionSize);
+
+        auto readSize =
+            ::read(panelFileDescriptor, versionBuffer.data(), versionSize);
+
+        if (versionSize == readSize)
+        {
+            isReadSuccessful = true;
+        }
+        else
+        {
+            std::cerr << "Failed to read device [" << devPath
+                      << "], number of bytes read: " << readSize
+                      << ", errno: " << errno << std::endl;
+        }
+    }
+    catch (const std::exception& ex)
+    {
+        std::cerr << "Failed device access, for the device [" << devPath
+                  << "], address: " << unsigned(devAddress)
+                  << ", error: " << ex.what() << std::endl;
+    }
+
+    if (panelFileDescriptor != -1)
+    {
+        if (close(panelFileDescriptor) == -1)
+        {
+            std::cerr << "close system call failed for the device [" << devPath
+                      << "], with error code: " << std::strerror(errno)
+                      << std::endl;
+        }
+    }
+    return isReadSuccessful;
+}
+
+bool enableDeviceI2cAccess(const panel::types::GpioInfoMap& gpioInfo,
+                           const std::string& devPath,
+                           const uint8_t& devAddress) noexcept
+{
+    try
+    {
+        if (gpioInfo.find("gpioPresence") != gpioInfo.end() &&
+            gpioInfo.find("setGpio") != gpioInfo.end())
+        {
+            bool errorReadingPin = false;
+
+            bool presenceValue = isDevicePresent(
+                std::get<0>(gpioInfo.at("gpioPresence")),
+                std::get<1>(gpioInfo.at("gpioPresence")), errorReadingPin);
+
+            if (presenceValue || errorReadingPin)
+            {
+                bool setGpioSuccess =
+                    setGpioPin(std::get<0>(gpioInfo.at("setGpio")),
+                               std::get<1>(gpioInfo.at("setGpio")));
+                if (setGpioSuccess && !errorReadingPin)
+                {
+                    return true;
+                }
+
+                // Read microcontroller firmware version
+                return isDeviceAccessible(devPath, devAddress);
+            }
+        }
+    }
+    catch (const std::exception& ex)
+    {
+        std::cerr
+            << "Error occured while enabling gpio line for the device path: "
+            << devPath << ", address: " << devAddress << ", error " << ex.what()
+            << std::endl;
+    }
+
+    // Failed to check panel presence
+    return false;
 }
 
 } // namespace utils
--- a/include/const.hpp
+++ b/include/const.hpp
@@ -129,5 +129,7 @@ static const std::string defaultHexWordValue = "00000000";
 
 // Maximum valid instance IDs' which can be created by libpldm
 static constexpr auto PLDM_INST_ID_MAX = 32;
+
+static constexpr auto RETRY_COUNT = 3;
 } // namespace constants
 } // namespace panel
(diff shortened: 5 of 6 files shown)


include/utils.hpp
```diff
@@ -40,6 +40,29 @@ static types::PanelDataMap lcdDataMap = {
      {constants::fujiLcdDevPath, constants::devAddr,
       constants::fujiLcdDbusObj}}};
 
+static types::LcdPanelGpioDataMap lcdGpioDataMap = {
```
souvik1914581: Should this be `static const` ?

include/utils.hpp
```diff
@@ -261,5 +284,61 @@ void sortPels(types::GetManagedObjects& listOfPels);
...
+ *
+ * @param gpioPinName - GPIO pin name.
+ * @param pinValue - GPIO pin value.
+ * @param errorReadingPin -  Falg to set in case of any error occured while
+ * reading the gpio pin.
+ *
+ * @return - Returns true if gpio pin is enabled, false otherwise.
+ */
+bool checkPresencePin(const std::string& gpioPinName, const uint8_t& pinValue,
+                      [[maybe_unused]] bool& errorReadingPin) noexcept;
```
souvik1914581: Why `[[maybe_unused]]`?

include/utils.hpp
```diff
@@ -261,5 +284,61 @@ void sortPels(types::GetManagedObjects& listOfPels);
 void filterPel(const types::GetManagedObjects& listOfPels,
                types::PelPathAndSRCList& finalListOFPELs);
 
+/**
+ * @brief API to check GPIO pin is enabled.
+ *
+ * This API checks whether the gpio presence pin is enabled or not.
+ *
+ * @param gpioPinName - GPIO pin name.
```
souvik1914581: nit: `@param` missing `[in/out]`

include/utils.hpp
```diff
@@ -261,5 +284,61 @@ void sortPels(types::GetManagedObjects& listOfPels);
...
+ * @param pinValue - GPIO pin value.
+ * @param errorReadingPin -  Falg to set in case of any error occured while
```
souvik1914581: typo: `Falg`

include/utils.hpp
```diff
@@ -261,5 +284,61 @@ void sortPels(types::GetManagedObjects& listOfPels);
...
+ * firmware version, if succeds then it considered as i2c line is enabled for
+ * the given device.
+ *
+ * @param gpioInfo - Map containing gpio presence and output pin information.
+ * @param devPath - Device I2c path.
+ * @param devAddress - Device address.
+ *
+ * @return - true if able to access the device, false otherwise.
+ */
+bool checkAndEnableGpioLine(panel::types::GpioInfoMap gpioInfo,
```
souvik1914581: `gpioInfo` can be passed as `const` reference?

meson.build
```diff
@@ -17,6 +17,11 @@ project(
 systemd = dependency('systemd')
 sdbusplus = dependency('sdbusplus')
 phosphor_dbus_interfaces = dependency('phosphor-dbus-interfaces')
+libgpiodcxx = dependency(
```
souvik1914581, line 20: Why are `meson` changes being bundled in with this PR?

src/panel_app_main.cpp
```diff
@@ -170,8 +172,23 @@ int main(int, char**)
...
+                        panel::utils::lcdGpioDataMap[imValue], lcdDevPath,
+                        lcdDevAddr))
+                {
+                    std::cout << "non everest system " << std::endl;
+                    lcdPanel->setTransportKey(true);
+                }
+            }
+            else
+            {
+                std::cout << "its everest system " << std::endl;
```
souvik1914581: 1. Is this trace needed?
2. How can we say "its everest system"? Hitting this `else` block just means that the `imValue` is not found in the map.

src/panel_app_main.cpp
```diff
@@ -170,8 +172,23 @@ int main(int, char**)
...
-            lcdPanel->setTransportKey(
-                panel::utils::getLcdPanelPresentProperty(imValue));
+            if (panel::utils::lcdGpioDataMap.find(imValue) !=
+                panel::utils::lcdGpioDataMap.end())
+            {
+                if (panel::utils::checkAndEnableGpioLine(
+                        panel::utils::lcdGpioDataMap[imValue], lcdDevPath,
+                        lcdDevAddr))
+                {
+                    std::cout << "non everest system " << std::endl;
```
souvik1914581: 1. Is this trace needed?

2. How can we say "non everest system"? Hitting this else block just means that the `imValue` is found in the map.

src/panel_app_main.cpp
```diff
@@ -170,8 +172,23 @@ int main(int, char**)
              * change from false to true; but the transport key is still
              * true(unchanged). To maintain data accuracy get the "Present"
              * property from dbus and set the transport key again.*/
-            lcdPanel->setTransportKey(
-                panel::utils::getLcdPanelPresentProperty(imValue));
+            if (panel::utils::lcdGpioDataMap.find(imValue) !=
+                panel::utils::lcdGpioDataMap.end())
+            {
+                if (panel::utils::checkAndEnableGpioLine(
```
souvik1914581: What happens if this `checkAndEnableGpioLine` fails? Should we continue further with the flow without setting transport key?

src/panel_app_main.cpp
```diff
@@ -4,6 +4,8 @@
 #include "const.hpp"
 #include "utils.hpp"
 
+#include <unistd.h>
```
souvik1914581, line 7: Why is this header required?

src/utils.cpp
```diff
@@ -442,5 +451,163 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
     return finalListOfFPELs;
 }
 
+bool checkPresencePin(const std::string& gpioPinName, const uint8_t& pinValue,
+                      [[maybe_unused]] bool& errorReadingPin) noexcept
```
souvik1914581: Is `[[maybe_unused]]` needed?

src/utils.cpp
```diff
@@ -442,5 +451,163 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+{
+    if (gpioPinName.empty())
```
souvik1914581, line 456: Should we check validity of `pinValue` as well? It should be either 0 or 1.

src/utils.cpp
```diff
@@ -442,5 +451,163 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+
+            l_presenceLine.request({"Read the presence line",
+                                    gpiod::line_request::DIRECTION_INPUT, 0});
+            errorReadingPin = false;
+            return (l_presenceLine.get_value() == pinValue);
+        }
+        catch (const std::exception& ex)
+        {
+            errorReadingPin = true;
+            std::cerr << "Error occured while checking gpio presence pin value "
```
souvik1914581: Retry trace below can be appended to this trace?

src/utils.cpp
```diff
@@ -442,5 +451,163 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+                         "for the pin : "
+                      << gpioPinName << ", error: " << ex.what() << std::endl;
+        }
+
+        std::cout << "GPIO presence pin, Retry count : " << retries
+                  << std::endl;
+        std::this_thread::sleep_for(1s);
```
souvik1914581: Have we tested this sleep time of 1s? In worst case, this API will block panel app initialization for 3 seconds.

src/utils.cpp
```diff
@@ -442,5 +451,163 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
 
+bool checkPresencePin(const std::string& gpioPinName, const uint8_t& pinValue,
+                      [[maybe_unused]] bool& errorReadingPin) noexcept
+{
+    if (gpioPinName.empty())
+    {
+        return false;
+    }
+
+    auto retries = 3;
```
souvik1914581: Should retries be passed as an argument so that caller can decide on the number of retries?

src/utils.cpp
```diff
@@ -442,5 +451,163 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+
+bool setGpioPin(const std::string& gpioPinName,
+                const uint8_t& pinValue) noexcept
+{
+    if (gpioPinName.empty())
+    {
+        return false;
+    }
+
+    auto retries = 3;
```
souvik1914581: Should retries be passed as an argument so that caller can decide on the number of retries?

src/utils.cpp
```diff
@@ -442,5 +451,163 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+                -1) // access failure
+            {
+                panel::types::Binary versionBuffer;
+                const size_t versionSize = 6;
+                versionBuffer.resize(versionSize);
+
+                auto readSize = ::read(panelFileDescriptor,
+                                       versionBuffer.data(), versionSize);
+
+                readSuccessful = (versionSize == readSize) ? true : false;
```
souvik1914581: This can simply be `readSuccessful = (versionSize == readSize)`

souvik1914581 (changes requested):

include/utils.hpp
branupama: done

include/utils.hpp
branupama: done

include/utils.hpp
branupama: done

include/utils.hpp
branupama: not required, removed.

include/utils.hpp
branupama: done

meson.build
branupama: gpio library dependency has been included in this PR, so need changes in meson to support this.

src/panel_app_main.cpp
branupama: not required, removed

src/panel_app_main.cpp
branupama: log is not required , removed.

src/panel_app_main.cpp
branupama: yes will continue further execution without transport, same as early implementation, where no impact on the transport but continues execution if device present property is not set.
gpio line will be detected with polling feature.

src/panel_app_main.cpp
branupama: trace not required. removed.

src/utils.cpp
branupama: removed

src/utils.cpp
branupama: I feel not required. with any invalid value call will fail.

src/utils.cpp
branupama: not required I guess, with buffer we are just trying again.
As part of polling device will be detected if inserted again.

src/utils.cpp
branupama: the common trace below should be enough.

src/utils.cpp
branupama: it wont block the panel, if the api is success at any iteration, API will  return immediately.

src/utils.cpp
branupama: I feel not required, just giving some time initially to retry looking for the device.
Its first time attempt to detect the device,

src/utils.cpp
branupama: done

include/utils.hpp
```diff
@@ -40,6 +40,29 @@ static types::PanelDataMap lcdDataMap = {
      {constants::fujiLcdDevPath, constants::devAddr,
       constants::fujiLcdDbusObj}}};
 
+static const types::LcdPanelGpioDataMap lcdGpioDataMap = {
```
jinuthomas: since all the lines are same under the IM , why don't we make only one and attach it to the different series , so if the IM supports enablement , the lines for both presence and enable should be the same.

include/utils.hpp
```diff
@@ -261,5 +284,62 @@ void sortPels(types::GetManagedObjects& listOfPels);
...
+ * This API checks whether the gpio presence pin is enabled or not.
+ *
+ * @param[in] gpioPinName - GPIO pin name.
+ * @param[in] pinValue - GPIO pin value.
+ * @param[in,out] errorReadingPin -  Flag to set in case of any error occured
+ * while reading the gpio pin.
+ *
+ * @return - Returns true if gpio pin is enabled, false otherwise.
+ */
+bool checkPresencePin(const std::string& gpioPinName, const uint8_t& pinValue,
```
jinuthomas: this should be 
bool  isDevicePresent(gpioPinName, expectedPinState)

include/utils.hpp
```diff
@@ -261,5 +284,62 @@ void sortPels(types::GetManagedObjects& listOfPels);
...
+ * the given device.
+ *
+ * @param gpioInfo[in] - Map containing gpio presence and output pin
+ * information.
+ * @param devPath[in] - Device I2c path.
+ * @param devAddress[in] - Device address.
+ *
+ * @return - true if able to access the device, false otherwise.
+ */
+bool checkAndEnableGpioLine(const panel::types::GpioInfoMap& gpioInfo,
```
jinuthomas: this will be
enableDeviceIicAccess()

include/utils.hpp
```diff
@@ -261,5 +284,62 @@ void sortPels(types::GetManagedObjects& listOfPels);
...
+
+/**
+ * @brief API to check microcontroller firmware version.
+ *
+ * @param[in] devPath - Device path.
+ * @param[in] devAddress - Device address.
+ *
+ * @return - true if able to read the firmware version, false otherwise.
+ */
+bool canReadFirmwareVersion(const std::string& devPath,
```
jinuthomas: this would be 
isDeviceAccessable()

src/panel_app_main.cpp
```diff
@@ -170,8 +172,21 @@ int main(int, char**)
...
+                if (panel::utils::checkAndEnableGpioLine(
+                        panel::utils::lcdGpioDataMap.at(imValue), lcdDevPath,
+                        lcdDevAddr))
+                {
+                    lcdPanel->setTransportKey(true);
+                }
+            }
+            else
+            {
+                lcdPanel->setTransportKey(
```
jinuthomas, line 189: set false here ?

src/utils.cpp
jinuthomas: why use literals and. not std::chrono::seconds(1)

src/utils.cpp
```diff
@@ -442,5 +449,162 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+    catch (const std::exception& ex)
+    {
+        std::cerr << "Failed to read firmware version for the device path: "
+                  << devPath << ", address:" << unsigned(devAddress)
+                  << ", error: " << ex.what() << std::endl;
+    }
+
+    if (panelFileDescriptor != -1)
+    {
+        close(panelFileDescriptor);
```
jinuthomas: could you check and log the return of close

src/utils.cpp
```diff
@@ -442,5 +449,162 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+            -1) // open failure
+        {
+            if (ioctl(panelFileDescriptor, I2C_SLAVE, devAddress) !=
+                -1) // access failure
+            {
+                panel::types::Binary versionBuffer;
+                const size_t versionSize = 6;
+                versionBuffer.resize(versionSize);
+
+                auto readSize = ::read(panelFileDescriptor,
```
jinuthomas: why is readSize not traced on failure ?

src/utils.cpp
```diff
@@ -442,5 +449,162 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+                      << gpioPinName << ", error: " << ex.what() << std::endl;
+        }
+
+        std::cout << "Set GPIO pin, Retry count : " << retries << std::endl;
+        std::this_thread::sleep_for(1s);
+    }
+    return false;
+}
+
+bool canReadFirmwareVersion(const std::string& devPath,
```
jinuthomas: whenever using linux system calls please trace the errno for debug

src/utils.cpp
```diff
@@ -442,5 +449,162 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+                auto readSize = ::read(panelFileDescriptor,
+                                       versionBuffer.data(), versionSize);
+
+                readSuccessful = (versionSize == readSize);
+            }
+        }
+    }
+    catch (const std::exception& ex)
+    {
+        std::cerr << "Failed to read firmware version for the device path: "
```
jinuthomas: failed device access will be generic

jinuthomas (changes requested):

include/utils.hpp
branupama: done

include/utils.hpp
branupama: updated with enableDeviceI2cAccess, as I feel its more readable.

include/utils.hpp
branupama: done

include/utils.hpp
branupama: done

src/panel_app_main.cpp
branupama: no, for the Device not found in the map types::LcdPanelGpioDataMap lcdGpioDataMap, should call setTransportKey() based on the dbus 'Present' property value.

src/utils.cpp
branupama: updated.

src/utils.cpp
branupama: added

src/utils.cpp
branupama: updated

src/utils.cpp
branupama: done

src/utils.cpp
branupama: updated

include/utils.hpp
jinuthomas: why map and not a vector of IM values ?

src/utils.cpp
jinuthomas: use a define to say retry_count as 3 and use it here

src/utils.cpp
```diff
@@ -442,5 +449,183 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+        }
+        catch (const std::exception& ex)
+        {
+            errorReadingPin = true;
+            std::cerr << "Error occured while checking gpio presence pin value "
+                         "for the pin : "
+                      << gpioPinName << ", error: " << ex.what() << std::endl;
+        }
+
+        std::cout << "GPIO presence pin, Retry count : " << retries
```
jinuthomas: Do we need to trace all the retry ?

src/utils.cpp
```diff
@@ -442,5 +449,183 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+                l_outputLine.request({"Panel App Action",
+                                      ::gpiod::line_request::DIRECTION_OUTPUT,
+                                      0},
+                                     pinValue);
+                return true;
+            }
+        }
+        catch (const std::exception& ex)
+        {
+            std::cout << "Error occured while enabling gpio line: "
```
jinuthomas: why is this out and not cerr

src/utils.cpp
```diff
@@ -442,5 +449,183 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+                      << gpioPinName << ", error: " << ex.what() << std::endl;
+        }
+
+        std::cout << "Set GPIO pin, Retry count : " << retries << std::endl;
```
jinuthomas: Is this trace needed ?

src/utils.cpp
jinuthomas: I would not suggest any trace here as in failures you get three lines of maybe same or different strings

src/utils.cpp
jinuthomas: I would not have this here.

src/utils.cpp
```diff
@@ -442,5 +449,183 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+            std::cerr << "open system call failed for the device [" << devPath
+                      << "], error code: " << std::strerror(errno) << std::endl;
+            return isReadSuccessful;
+        }
+        if (ioctl(panelFileDescriptor, I2C_SLAVE, devAddress) ==
+            -1) // access failure
+        {
+            std::cerr << "ioctl call failed for the device [" << devPath
+                      << "], error code: " << std::strerror(errno) << std::endl;
+            return isReadSuccessful;
```
jinuthomas: close not called here

include/utils.hpp
branupama: searching in map with direct IM as keyword is more efficient than in the list/vector.

src/utils.cpp
branupama: done

src/utils.cpp
branupama: summarised the log at the end .

src/utils.cpp
branupama: updated

src/utils.cpp
branupama: summarised the log at the end .

src/utils.cpp
branupama: summarised the log at the end .

src/utils.cpp
branupama: removed, kept summarised log at the end.

include/utils.hpp
```diff
@@ -261,5 +274,63 @@ void sortPels(types::GetManagedObjects& listOfPels);
...
+ *
+ * This API checkes for the gpio presence pin, if found enabled or in case of
+ * any error, will try to set the gpio output pin. If both reading input pin and
+ * enabling output pin is failed, will try to directly read the microcontroller
+ * firmware version, if succeds then it considered as i2c line is enabled for
+ * the given device.
+ *
+ * @param gpioInfo[in] - Map containing gpio presence and output pin
+ * information.
+ * @param devPath[in] - Device I2c path.
```
souvik1914581: nit: `i2c`

include/utils.hpp
```diff
@@ -261,5 +274,63 @@ void sortPels(types::GetManagedObjects& listOfPels);
...
+ * @param devAddress[in] - Device address.
+ *
+ * @return - true if able to access the device, false otherwise.
+ */
+bool enableDeviceI2cAccess(const panel::types::GpioInfoMap& gpioInfo,
+                           const std::string& devPath,
+                           const uint8_t& devAddress) noexcept;
+
+/**
+ * @brief API to check microcontroller firmware version.
```
souvik1914581: `API to check if device is accessible` ?

`This API checks if device is accessible by trying to read the microcontroller firmware version` ?

src/utils.cpp
```diff
@@ -442,5 +449,191 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+    {
+        try
+        {
+            gpiod::line l_outputLine = gpiod::find_line(gpioPinName);
+            if (l_outputLine)
+            {
+                l_outputLine.request({"Panel App Action",
+                                      ::gpiod::line_request::DIRECTION_OUTPUT,
+                                      0},
+                                     pinValue);
```
souvik1914581, line 521: What happens if we pass any `pinValue` other than `0` or `1` to this API?

src/utils.cpp
```diff
@@ -442,5 +449,191 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+        }
+        if (ioctl(panelFileDescriptor, I2C_SLAVE, devAddress) ==
+            -1) // access failure
+        {
+            throw std::runtime_error("ioctl call failed, error code: " +
+                                     std::string(std::strerror(errno)));
+        }
+
+        panel::types::Binary versionBuffer;
+        const size_t versionSize = 6;
```
souvik1914581: `constexpr size_t versionSize`

src/utils.cpp
```diff
@@ -442,5 +449,191 @@ types::PelPathAndSRCList geListOfPELsAndSRCs()
...
+        versionBuffer.resize(versionSize);
+
+        auto readSize =
+            ::read(panelFileDescriptor, versionBuffer.data(), versionSize);
+
+        if (versionSize == readSize)
+        {
+            isReadSuccessful = true;
```
souvik1914581, line 570: This API is trying to check if the device is accessible. If the number of bytes read is  not -1 but is less than expected number of bytes, should we declare that the device is not accessible?

It is not an error if `read()`  returns less than expected number of bytes. It means the device is accessible but the `read()` operation was interrupted.

include/utils.hpp
jinuthomas: I disagree

include/utils.hpp
```diff
@@ -261,5 +274,63 @@ void sortPels(types::GetManagedObjects& listOfPels);
...
+ * @brief API to check GPIO pin is enabled.
+ *
+ * This API checks whether the gpio presence pin is enabled or not.
+ *
+ * @param[in] gpioPinName - GPIO pin name.
+ * @param[in] expectedPinState - GPIO pin value to check.
+ * @param[in,out] errorReadingPin -  Flag to set in case of any error occured
+ * while reading the gpio pin.
+ *
+ * @return - Returns true if gpio pin is enabled, false otherwise.
```
jinuthomas: Returns true if device is present

include/utils.hpp
```diff
@@ -261,5 +274,63 @@ void sortPels(types::GetManagedObjects& listOfPels);
 void filterPel(const types::GetManagedObjects& listOfPels,
                types::PelPathAndSRCList& finalListOFPELs);
 
+/**
+ * @brief API to check GPIO pin is enabled.
```
jinuthomas: Name needs to change

include/utils.hpp
```diff
@@ -261,5 +274,63 @@ void sortPels(types::GetManagedObjects& listOfPels);
...
+ * @param gpioPinName[in] - GPIO pin name.
+ * @param pinValue[in] - GPIO pin value.
+ *
+ * @return - Returns true if able to set gpio pin, false otherwise.
+ */
+bool setGpioPin(const std::string& gpioPinName,
+                const uint8_t& pinValue) noexcept;
+
+/**
+ * @brief API to enable gpio line for the given device.
```
jinuthomas: enable gpio access

include/utils.hpp
branupama: Agree, with small data set of vector is stores in cache location, hence faster than the map to access the data.
Updated the code.

include/utils.hpp
branupama: updated

include/utils.hpp
branupama: updated

include/utils.hpp
branupama: done

include/utils.hpp
branupama: updated

include/utils.hpp
branupama: done

src/utils.cpp
branupama: request will fails for invalid value. logs are captured in failure case already.

src/utils.cpp
branupama: done

src/utils.cpp
branupama: as we are reading from the device, either will read 6 bytes or call will fails.
so fine here to check read 6 bytes or not.

jinuthomas (approved):

PR #243, opened 2025-12-29
Prevent panel crash when IM is not found

branupama:
The panel application no longer crashes when the IM value cannot be
retrieved from D-Bus. A critical PEL is logged and the application
continues to run.

When IM is not supported, the panel state manager still maintains
system state and D-Bus APIs remain available for communication.

This commit also removes panel support for the Tacoma system.

Testing:
```
Tested the panel application on the Huygens Simics system, for which panel support
is not yet available.
* A critical PEL is logged for the missing IM.
* Using paneltool, I was able to view the enabled functions on the panel and
successfully change the mode to manual mode using function 2.
* The panel application continued to operate normally.

Tested the same changes on a P10 Rainier system.
* Observed that the LCD panel working correctly.
* LCD buttons were functional and able to execute functions displayed on the screen
without any issues.
```

### DIFF
--- a/src/panel_app_main.cpp
+++ b/src/panel_app_main.cpp
@@ -60,20 +60,16 @@ std::string getInputDevicePath(const std::string& imValue)
         return "/dev/input/by-path/platform-1e78a780.i2c-event-joystick";
     }
 
-    // default to tacoma
-    return "/dev/input/by-path/platform-1e78a080.i2c-bus-event-joystick";
+    return std::string{};
 }
 
 void getLcdDeviceData(std::string& lcdDevPath, uint8_t& lcdDevAddr,
                       std::string& lcdObjPath, const std::string& imValue)
 {
-    if (panel::utils::lcdDataMap.find(imValue) ==
-        panel::utils::lcdDataMap.end()) // assume the system is tacoma
-    {
-        lcdDevPath = panel::constants::tacomaLcdDevPath;
-        lcdDevAddr = panel::constants::devAddr;
-    }
-    else
+    lcdDevAddr = 0;
+
+    if (panel::utils::lcdDataMap.find(imValue) !=
+        panel::utils::lcdDataMap.end())
     {
         lcdDevPath =
             std::get<0>((panel::utils::lcdDataMap.find(imValue))->second);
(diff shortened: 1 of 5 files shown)


src/panel_app_main.cpp
```diff
@@ -98,10 +93,31 @@ int main(int, char**)
             server.add_interface("/com/ibm/panel_app", "com.ibm.panel");
 
         const std::string imValue = panel::utils::getSystemIM();
+        if (imValue.empty())
+        {
+            std::map<std::string, std::string> additionData{
```
souvik1914581: This map can be `const` or can be declared in-place in the `createPEL()` call

src/panel_app_main.cpp
```diff
@@ -98,10 +93,31 @@ int main(int, char**)
...
+                {"DESCRIPTION", "Failed read IM value from Dbus"}};
+            panel::utils::createPEL(
```
souvik1914581, line 252: nit: a line gap would make this more readable

src/panel_app_main.cpp
```diff
@@ -67,13 +67,8 @@ std::string getInputDevicePath(const std::string& imValue)
 void getLcdDeviceData(std::string& lcdDevPath, uint8_t& lcdDevAddr,
                       std::string& lcdObjPath, const std::string& imValue)
 {
-    if (panel::utils::lcdDataMap.find(imValue) ==
+    if (panel::utils::lcdDataMap.find(imValue) !=
```
souvik1914581, line 71: Some context behind this change would be helpful.

src/panel_app_main.cpp
```diff
@@ -98,10 +93,31 @@ int main(int, char**)
...
+                additionData);
+            return -1;
+        }
 
         std::string lcdDevPath{}, lcdObjPath{};
         uint8_t lcdDevAddr;
         getLcdDeviceData(lcdDevPath, lcdDevAddr, lcdObjPath, imValue);
+        if (lcdDevPath.empty() || lcdObjPath.empty())
+        {
+            std::map<std::string, std::string> additionData{
```
souvik1914581: This map can be `const` or can be declared in-place in the `createPEL()` call

src/transport.cpp
```diff
@@ -96,6 +96,14 @@ void Transport::panelI2CWrite(const types::Binary& buffer) const
                     failedErrno = errno;
                     sleep(1);
                     const std::string imValue = utils::getSystemIM();
+                    if (imValue.empty())
+                    {
+                        std::cerr << "I2C write failed, as failed to get IM "
+                                     "value form DBus."
```
souvik1914581: nit: "from"

src/utils.cpp
```diff
@@ -313,6 +313,13 @@ std::string getSystemIM()
 
 bool getLcdPanelPresentProperty(const std::string& imValue)
 {
+    if (lcdDataMap.find(imValue) == lcdDataMap.end())
+    {
+        std::cerr << "\n Panel info is missing for the given IM value."
```
souvik1914581: Why is the new line character needed at the beginning of this log message?

src/panel_app_main.cpp
branupama: done

src/panel_app_main.cpp
branupama: done

src/panel_app_main.cpp
branupama: done

src/transport.cpp
branupama: done

src/utils.cpp
branupama: removed new line

src/panel_app_main.cpp
branupama: this will only query if imValue found in lcdDataMap, I guess we can understand directly without any comment.

src/panel_app_main.cpp
```diff
@@ -101,133 +189,103 @@ int main(int, char**)
...
-                    baseObjPath, conn, basePanel, stateManager);
-
-                basePanelPresence->listenPanelPresence();
-            }
 
-            basePanel->setTransportKey(true);
-        }
-
-        // Listen to lcd panel presence always for both rainier and everest
         std::unique_ptr<panel::PanelPresence> presence;
```
SunnySrivastava1984: Do we need the variables at this scope?

SunnySrivastava1984 (changes requested):

src/panel_app_main.cpp
branupama: after the code update variable scope is kept same.

src/panel_app_main.cpp
souvik1914581: Earlier, if IM value was not found in `lcdDataMap`, we assumed system is Tacoma. 
With this change, `lcdDevPath` and `lcdDevAddr` would be left unchanged, should we set them to empty string?

src/panel_app_main.cpp
```diff
@@ -84,6 +78,156 @@ void getLcdDeviceData(std::string& lcdDevPath, uint8_t& lcdDevAddr,
...
+ * @param[in] lcdDevAddr - LCD i2c device address.
+ * @param[in, out] lcdPanel - Transport object.
+ * @param[in] stateManager - PanelStateManager object.
+ * @param[in] conn - Dbus connection.
+ * @param[out] lcdPanelPresence - PanelPresence object.
+ *
+ * @throw std::runtime_error, std::bad_alloc
+ */
+void setupLcdPanel(
+    const std::string& imValue, std::string& lcdObjPath,
```
souvik1914581, line 171: `lcdObjectPath` should be `const`?

src/panel_app_main.cpp
branupama: done

src/panel_app_main.cpp
branupama: PanelPresence class takes non-const value, so kept this variable without const to avoid the compilation error.

src/panel_app_main.cpp
souvik1914581: Ok

souvik1914581:
I think the commit message should be updated to reflect that setting up Base panel and LCD panel has been refactored and consolidated into respective APIs.

branupama:
> setting up Base panel and LCD panel has been refactored and consolidated into respective APIs.

Done

SunnySrivastava1984 (approved):

PR #248, opened 2026-06-30
Disable SRC functions when progress code is cleared

branupama:
Functions 11, 12, 13 were remaining enabled on the panel after the host reached running state because no action was taken when the clear display progress code was received. This caused SRC functions to remain visible without a corresponding PEL being logged.

The code changes made to fix the above issue by disabling functions 11–19 via the state manager and resetting the PEL listener's functionStateEnabled flag when the zero/clear progress code event is received.

### DIFF
--- a/include/bus_monitor.hpp
+++ b/include/bus_monitor.hpp
@@ -100,6 +100,24 @@ class PELListener
      */
     void listenPelEvents();
 
+    /**
+     * @brief Disables the function state flag for SRC-related functions.
+     *
+     * Sets the internal functionStateEnabled flag to false, preventing
+     * SRC functions (11–19) from being treated as active. This should
+     * be called when the panel display is cleared (e.g., on receiving
+     * clear progress code) to ensure SRC functions are not
+     * accessible until a new PEL is logged.
+     *
+     * @note This flag is also evaluated by the PEL event listener to
+     *       determine whether SRC functions should be re-enabled on the
+     *       next PEL event.
+     */
+    void DisableFunctionState()
+    {
+        functionStateEnabled = false;
+    }
+
   private:
     /* Callback to listen for PEL event log */
     void PELEventCallBack(sdbusplus::message_t& msg);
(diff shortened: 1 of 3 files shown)


include/bus_monitor.hpp
```diff
@@ -100,6 +101,14 @@ class PELListener
      */
     void listenPelEvents();
 
+    /**
+     * @brief disable function state on zero progress code event.
```
SunnySrivastava1984: This is a public API, how are you restricting it to be called only on zero progress code event?
Also add some more detail for the API and its use case.

include/bus_monitor.hpp
```diff
@@ -138,7 +147,7 @@ class PELListener
     std::shared_ptr<Transport> transport;
 
     /* Check if respective functions are enabled */
-    bool functionStateEnabled = false;
+    std::atomic<bool> functionStateEnabled{false};
```
SunnySrivastava1984: Any specific use case against this change?

SunnySrivastava1984 (changes requested):

include/bus_monitor.hpp
branupama: BootProgressCode calls DisableFunctionState() on PELListener, and PELListener itself also writes functionStateEnabled in its own callbacks. 
The variable is updated from two separate callback contexts. Marking it std::atomic<bool> makes the intent explicit and keeps it safe against if there any future threading changes.

include/bus_monitor.hpp
branupama: updated the doxygen

include/bus_monitor.hpp
SunnySrivastava1984: The flow is single threaded, in that case I don't see a race case happening.

> Marking it std::atomic makes the intent explicit and keeps it safe against if there any future threading changes.

1120 is in service pack phase, is there a possibility that we will move the code to multithreaded design in 11 series?

branupama: we are not moving to multithreading, reverted the changes.

SunnySrivastava1984 (approved):

SunnySrivastava1984 (approved):

PR #250, opened 2026-07-31
Disable SRC functions when progress code is cleared

branupama:
Functions 11, 12, 13 were remaining enabled on the panel after the host reached running state because no action was taken when the clear display progress code was received. This caused SRC functions to remain visible without any new PEL is being logged.

The code changes made to fix the above issue by disabling functions 11–19 via the state manager and resetting the PEL listener's functionStateEnabled flag when the zero/clear progress code event is received.

Also disabled functions 14–19 on each new progress code event to clear callout SRCs displayed from the previous PEL.

```
Test:
When Host reaches running state, functions 11, 12, 13 are disabled not show on the panel.

Induced critical PEL, observed that corresponding SRC functions 11-19 are enabled back on the panel.

```

### DIFF
--- a/include/bus_monitor.hpp
+++ b/include/bus_monitor.hpp
@@ -100,6 +100,24 @@ class PELListener
      */
     void listenPelEvents();
 
+    /**
+     * @brief Disables the function state flag for SRC-related functions.
+     *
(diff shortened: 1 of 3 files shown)


SunnySrivastava1984 (approved):
Approved. Cherry pick commit.

PR #251, opened 2026-08-12
Introduce main.cpp to support PST systems

branupama:
Introduce an entry point ` main.cpp` to support PST systems, replacing
the existing `panel_app_main.cpp`.
    
Key changes:
  - Created `src/main.cpp` to run the basic D-Bus service loop.
  - `lg2` from `phosphor-logging` is used for logging .
  - Pruned and updated  to compile only
    into the  executable. Removed all unused libraries, tests,
    and tools that are not required for the initial service boot.
    
Required modules will be integrated along with their respective changes
in future commits.
```
Test:
Loaded changes on the PST simics, observed the Panel system service is up and running
and com.ibm.PanelApp dbus service is hosting /com/ibm/panel_app object path.
```


include/const.hpp
```diff
@@ -9,6 +9,10 @@ namespace panel
 namespace constants
 {
 
+static constexpr auto panelService = "com.ibm.PanelApp";
+static constexpr auto panelInterface = "com.ibm.panel";
+static constexpr auto panelObjectPath = "/com/ibm/panel_app";
+
```
SunnySrivastava1984: Get rid of rest of the data.

SunnySrivastava1984 (changes requested):
Remove unused data from const.hpp. Rest looks ok.

include/const.hpp
branupama: done

SunnySrivastava1984 (approved):

PR #252, opened 2026-08-12
Fetch bmc role

branupama:


include/const.hpp
```diff
@@ -12,5 +12,18 @@ namespace constants
 static constexpr auto panelService = "com.ibm.PanelApp";
 static constexpr auto panelInterface = "com.ibm.panel";
 static constexpr auto panelObjectPath = "/com/ibm/panel_app";
+
+static constexpr auto redundancyService =
+    "xyz.openbmc_project.State.BMC.Redundancy";
+static constexpr auto redundancyObjectPath = "/xyz/openbmc_project/state/bmc0";
+static constexpr auto redundancyInterface =
+    "xyz.openbmc_project.State.BMC.Redundancy";
+static constexpr auto redundancyPropertyRole = "Role";
```
SunnySrivastava1984: Why it is redundancy property role? This itself is just propert role.

include/const.hpp
```diff
@@ -12,5 +12,18 @@ namespace constants
...
+static constexpr auto redundancyRolePassive =
+    "xyz.openbmc_project.State.BMC.Redundancy.Role.Passive";
```
SunnySrivastava1984, line 23: Same concern for rest as well. Why are they prefixed with redundancy?

src/utils.cpp
```diff
@@ -1,647 +1,40 @@
...
-        if (versionSize == readSize)
-        {
-            isReadSuccessful = true;
-        }
-        else
-        {
-            std::cerr << "Failed to read device [" << devPath
-                      << "], number of bytes read: " << readSize
-                      << ", errno: " << errno << std::endl;
+            lg2::error("Invalid or empty data received for BMC role from DBus, "
```
SunnySrivastava1984: While debugging, how it will be differentaited that we got empty data or the variant type was invalid?

src/utils.cpp
```diff
@@ -1,647 +1,40 @@
...
-    catch (const sdbusplus::exception_t& e)
-    {
-        std::cerr << "Error in invoking D-Bus logging create interface to "
-                     "register PEL";
-    }
-}
 
-std::string getService(sdbusplus::bus_t& bus, const std::string& path,
-                       const std::string& interface)
+std::string getBmcRedundancyRole() noexcept
```
SunnySrivastava1984: Just an opinion, since we are starting in a fresh.
Should we maintain only one file for utility methods "utils.hpp" as we praactise in another repos. This takes way the overhead of declaring and defining at thwo different palces.

src/main.cpp
```diff
@@ -24,6 +26,10 @@ int main()
             server.add_interface(panel::constants::panelObjectPath,
                                  panel::constants::panelInterface);
 
+        // Determine BMC Redundancy Role
+        const std::string bmcRole = panel::utils::getBmcRedundancyRole();
+        lg2::info("Determined BMC Redundancy Role: {ROLE}.", "ROLE", bmcRole);
```
SunnySrivastava1984: Do we need this info here?
I guess once panel starts in a mode, we can have 1 info log there stating that based on role "xyz" panel is starting in "abc" mode?

SunnySrivastava1984: Just to keep in mind that there will be many instantiation done in due process.
We should not put everything in main itself instead break them into meaningful small chunks of functions.

branupama: done

src/utils.cpp
branupama: made changes to use single utils.hpp

include/const.hpp
branupama: updated

include/const.hpp
branupama: as the value contains the redundancy in it used the same convention. updated the name

src/utils.cpp
branupama: added logs for all scenarios.

include/const.hpp
branupama: removed the redundancy name

RekhaAparna01 (approved):

include/utils.hpp
```diff
@@ -75,264 +34,92 @@ T readBusProperty(const std::string& service, const std::string& object,
...
 
-/** @brief Convert byte vector to hex string.
- * @param[in] val - byte vector that needs conversion
- * @return hex string
- */
-std::string binaryToHexString(const types::Binary& val);
-
 /**
- * @brief Api to create PEL.
+ * @brief Determine the redundancy role of the BMC from D-Bus.
```
SunnySrivastava1984, line 44: What is meant by redundancy role? Do we have a redundancy specific BMC role?

include/utils.hpp
```diff
@@ -75,264 +34,92 @@ T readBusProperty(const std::string& service, const std::string& object,
...
- * @param[in] errIntf - Error Interface
- * @param[in] sev -  panel::constants::Severity of Error
- * @param[in] additionalData - Information of PEL
- */
-void createPEL(const std::string& errIntf, const std::string& sev,
-               const std::map<std::string, std::string>& additionalData);
-
-/**
- * @brief Api to get service.
+ * This API queries the BMC redundancy service to fetch the current Role
```
SunnySrivastava1984, line 46: If there is a dependency on this service then do we need to have the service dependency set?

include/utils.hpp
```diff
@@ -75,264 +34,92 @@ T readBusProperty(const std::string& service, const std::string& object,
...
+ *         exists for the supplied role.
  */
-bool enableDeviceI2cAccess(const panel::types::GpioInfoMap& gpioInfo,
-                           const std::string& devPath,
-                           const uint8_t& devAddress) noexcept;
+inline std::string getModeBasedOnBmcRole(const std::string& role)
+{
+    if (role.compare(constants::rolePassive) == 0)
+    {
+        return "Passive";
```
SunnySrivastava1984, line 110: Why do we need this function and hard coded strings just for logging?

src/main.cpp
SunnySrivastava1984: I can still see the log here. Then how is it done?

src/main.cpp
```diff
@@ -24,6 +26,13 @@ int main()
             server.add_interface(panel::constants::panelObjectPath,
                                  panel::constants::panelInterface);
 
+        // Determine BMC Redundancy Role
+        const std::string bmcRole = panel::utils::getBmcRedundancyRole();
+        lg2::info(
```
SunnySrivastava1984, line 31: We are still having log here?

include/utils.hpp
branupama: updated

include/utils.hpp
branupama: not, required.
On panel app start, query this service. if service is not  running/available, will conside role as unknow.
later when service is up, with the listener role will be updated with new role.

include/utils.hpp
branupama: instead of having this logic where we need to display, api will give the value in string format, can be used while logging in future for other apis as well.

src/main.cpp
branupama: as suggested early added a log, to diaplay role on which panel is starting.

